PR #243, opened 2026-04-08
fix(docker): add proxy support and fix missing runtime libs for webui-server mode

freekingxx:
## Problem

The existing Dockerfile has two issues that prevent the Docker image from being built or run in certain environments:

### 1. Build fails behind HTTP/HTTPS proxy

`corepack enable && corepack install` are chained in a single `RUN` command. Because `corepack enable` runs first (before any proxy env var is available), `corepack install` tries to download pnpm directly from `registry.npmjs.org` without going through the proxy — and fails:

```
Internal Error: Error when performing the request to https://registry.npmjs.org/pnpm/-/pnpm-10.13.1.tgz
```

The same problem applies to `apt-get` in the backend stage: some intercepting proxies return HTTP 502 for plain HTTP requests, requiring the apt sources to be switched to HTTPS first.

### 2. Container crashes immediately at runtime

The binary exits with:

```
cchv-server: error while loading shared libraries: libgdk-3.so.0: cannot open shared object file: No such file or directory
```

Although the binary runs in headless `--serve` mode (no GUI), **Tauri links against webkit2gtk/gtk3 at compile time**. These shared libraries must therefore be present in the runtime image. The previous Stage 3 only installed `ca-certificates` and `curl`, leaving all GTK/WebKit `.so` files missing.

## Changes

| Stage | Change | Reason |
|-------|--------|--------|
| All | `ARG PROXY_URL` + `ENV http_proxy/https_proxy` | Allow proxy-aware downloads |
| Frontend | Split `corepack enable && corepack install` into separate `RUN` steps | Proxy env must be set before `corepack install` fetches pnpm |
| Backend | `rust:1.82-bookworm` → `rust:1-bookworm` | Unpin stale patch version; use floating minor |
| Backend | `sed` apt sources to HTTPS before `apt-get update` | Avoid 502 from MITM proxies on HTTP |
| Runtime | Add `libwebkit2gtk-4.1-0 libgtk-3-0 libjavascriptcoregtk-4.1-0` | Satisfy Tauri's compile-time shared-library dependencies at runtime |
| Runtime | `ENV http_proxy= https_proxy=` at end of stage | Ensure proxy vars are not baked into the final image |

## Testing

```bash
# Build with proxy
docker build --build-arg PROXY_URL=http://<proxy-host>:<port> -t cchv-test .

# Build without proxy (unchanged behaviour)
docker build -t cchv-test .

# Verify health endpoint
docker run --rm -d -p 3727:3727 cchv-test
curl http://localhost:3727/health
# → {"status":"ok"}
```

Tested on: Debian Bookworm / WSL2 / Docker Engine 29.1.2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,8 @@
 
 # ── Stage 1: Build frontend ──────────────────────────────────────────
 FROM node:20-slim AS frontend
+ARG PROXY_URL
+ENV http_proxy=${PROXY_URL} https_proxy=${PROXY_URL}
 RUN corepack enable
 WORKDIR /app
 COPY package.json pnpm-lock.yaml ./
@@ -13,8 +15,13 @@ RUN pnpm exec tsc --build . && pnpm exec vite build
 
 # ── Stage 2: Build Rust server binary (with embedded frontend) ──────
 FROM rust:1-bookworm AS backend
-RUN apt-get update && apt-get install -y --no-install-recommends \
-    libwebkit2gtk-4.1-dev libappindicator3-dev librsvg2-dev patchelf \
+ARG PROXY_URL
+ENV http_proxy=${PROXY_URL} https_proxy=${PROXY_URL}
+# Force HTTPS for apt sources to avoid proxy 502 on HTTP
+RUN sed -i 's|http://deb.debian.org|https://deb.debian.org|g' /etc/apt/sources.list.d/debian.sources \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+       libwebkit2gtk-4.1-dev libappindicator3-dev librsvg2-dev patchelf \
     && rm -rf /var/lib/apt/lists/*
 WORKDIR /app
 COPY src-tauri/ src-tauri/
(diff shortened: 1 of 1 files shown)


jhlee0409:
@freekingxx Thanks for the contribution! 🙏

This PR currently has merge conflicts with `develop` because several of the fixes here have already been addressed in recent commits:

- **`corepack install` ordering** → fixed in 58af5a5
- **`rust:1.82` → flexible version** → updated to `rust:1-slim-bookworm` in 43e9794
- **Runtime libs (`libgtk-3-0`, `libwebkit2gtk`)** → added in bd1d35d

The **proxy support (`PROXY_URL` ARG)** is the new and valuable part of this PR.

Could you rebase this branch onto `develop` and keep only the proxy-related changes? That should resolve the conflicts and make this ready to merge.

```bash
git fetch upstream
git rebase upstream/develop
# resolve conflicts by keeping develop's version for the already-fixed parts,
# and keeping your proxy additions
```

Let me know if you need any help with the rebase!

jhlee0409:
Rebased onto `develop`, resolved conflicts (kept existing fixes + your proxy additions), and merged. Thanks for the contribution! 🎉

PR #268, opened 2026-04-18
fix: render context menus in a portal so they anchor to the cursor

greenbritainclub-ux:
## Summary

The right-click context menus on session and project items were rendering detached from the cursor — offset horizontally, sometimes clipped against the pane edge.

**Root cause:** `SessionList` uses `react-window`'s `FixedSizeList`, whose outer scroll container has `will-change: transform`. Per CSS spec, `will-change: transform` (like `transform` itself) creates a new containing block for descendants with `position: fixed`. The context menu is rendered as a descendant of that scroll container, so its `fixed` coordinates were being resolved relative to the virtualized scroll box instead of the viewport.

**Fix:** render both context menus via `createPortal(..., document.body)` so they escape the transformed ancestor and anchor to `clientX`/`clientY` as intended.

Also switched `ProjectContextMenu`'s off-screen adjustment from `useEffect` to `useLayoutEffect` and added a `Math.max(8, …)` clamp, matching the logic already in `SessionContextMenu` — prevents a single-frame flash at the wrong position and keeps the menu from popping off the top/left edges.

## Changes

- `src/components/SessionItem/components/SessionContextMenu.tsx` — wrap return in `createPortal`
- `src/components/ProjectContextMenu.tsx` — wrap return in `createPortal`, `useLayoutEffect` + edge clamp

## Test plan

- [x] `pnpm tsc --build .` — clean
- [x] `pnpm lint` — clean (pre-existing unrelated warning in `GlobalSearchModal.tsx`)
- [x] `pnpm vitest run` — 735/735 pass
- [ ] Manual: right-click a session item inside the virtualized list — menu should appear directly under the cursor, not floating off to the right
- [ ] Manual: right-click near the viewport edge — menu should slide back in-bounds as before
- [ ] Manual: right-click a project in `ProjectTree` — menu should anchor correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved context menu positioning and rendering behavior for more reliable UI placement across components.

### DIFF
--- a/src/components/ProjectContextMenu.tsx
+++ b/src/components/ProjectContextMenu.tsx
@@ -1,5 +1,6 @@
 // src/components/ProjectContextMenu.tsx
-import React, { useState, useEffect, useRef } from "react";
+import React, { useState, useEffect, useLayoutEffect, useRef } from "react";
+import { createPortal } from "react-dom";
 import { EyeOff, Eye, Copy } from "lucide-react";
 import { useTranslation } from "react-i18next";
 import { toast } from "sonner";
@@ -50,7 +51,7 @@ export const ProjectContextMenu: React.FC<ProjectContextMenuProps> = ({
   }, [onClose]);
 
   // Adjust position if menu would go off-screen
-  useEffect(() => {
+  useLayoutEffect(() => {
     if (menuRef.current) {
       const rect = menuRef.current.getBoundingClientRect();
       const windowWidth = window.innerWidth;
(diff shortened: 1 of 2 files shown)


jhlee0409:
Thanks for the contribution — the diagnosis (CSS containing block from `will-change: transform` on the virtualized list) is spot-on, and the fix is clean and minimal. Verified locally: `tsc`, `lint`, and `vitest` (735/735) all pass on the branch, and the outside-click logic still works correctly since it uses `node.contains()` on the DOM tree rather than the React tree.

Appreciate the detailed write-up.

PR #273, opened 2026-04-20
fix(cursor): preserve bubble timestamps in search results (#266)

### LINKED ISSUE #266: Cursor Invalid Date and  "Session not found" when opening a session from the search
Hi there,

I am trying to troubleshoot why I am getting "Session not found" from Cursor, if Claude Code History Viewer already found the session in the search.

I also noticed that it is showing as "Invalid Date".

Image

isimple4:
Fixes #266

## Summary

Ensure Cursor search results carry the original bubble creation timestamp so the global search modal can render a valid date instead of `Invalid Date`.

## Root cause

Cursor bubbles already store a creation timestamp in `createdAt`, but the provider conversion path built search/message results with an empty `timestamp` field. The global search modal formats that field directly, so empty timestamps surfaced as `Invalid Date`.

## Changes

- extract Cursor bubble timestamps from `createdAt` first, with fallback support for string or millisecond `timestamp` values
- pass the resolved timestamp through both user and assistant Cursor bubble conversion paths
- add regression tests covering user and assistant bubble timestamp mapping

## Test plan

- [x] `cargo test --manifest-path src-tauri/Cargo.toml providers::cursor::tests::test_convert_`
- [x] `cargo fmt --manifest-path src-tauri/Cargo.toml --all -- --check`

## Notes

This PR only fixes the `Invalid Date` portion of #266. It does not address the separate Cursor search navigation issue that can still surface `Session not found`.

## Screenshots
image

## Release Notes

* **Bug Fixes**
  * Fixed missing message timestamps. Messages now properly extract and display creation timestamps in ISO 8601 format, with improved support for multiple timestamp sources including numeric millisecond values.

* **Tests**
  * Enhanced test coverage with new cases validating timestamp extraction and conversion for different input formats.


isimple4:
> Right now any non-empty string wins before millisecond conversion. If Cursor stores "1700000000000" or another malformed string, it will still flow into ClaudeMessage.timestamp and fail downstream RFC3339 parsing, so search can still show an invalid date.

In the Cursor bubble data I sampled locally, `createdAt` is consistently an RFC3339/ISO string

isimple4:
I checked the failing `Lint & Format (pull_request)` job.

This is unrelated to the Cursor timestamp change in this PR. The failure comes from `cargo clippy -D warnings` on existing warnings in other files from the merged PR ref, specifically `src/commands/stats.rs` and `src/providers/codex.rs`.

The change in this PR is limited to `src-tauri/src/providers/cursor.rs`, so this looks like an unrelated CI failure outside the scope of the `Invalid Date` fix.

PR #277, opened 2026-04-21
feat: follow symlinks at depth 1 in scan_projects

### LINKED ISSUE #259: Feature request: opt-in support for symlinked project directories
## Use case                                                                                                                                                            
                                                                    
I have multiple macOS user accounts on the same machine and share Claude Code sessions between them using symlinks — I put the actual project folders in `/Users/Shared/.claude/projects/` and symlink them into each account's `~/.claude/projects/`. Pretty standard macOS pattern for cross-account sharing.

The problem is CCHV silently skips those symlinked directories, so the shared sessions never show up.

## Why it doesn't work today

`scan_projects` filters on `file_type().is_dir()`, which returns `false` for symlinks. I noticed this is intentional — symlinks are explicitly blocked as a security measure during directory traversal, which makes sense as a default.

## What I'd suggest

Would you be open to an **opt-in toggle** in settings (something like "Follow symlinks in projects folder")? Off by default so nothing changes for existing users. When enabled, it would follow symlinks at depth 1 only (project directories), not deeper — which keeps the risk minimal since it can't escape the `projects/` boundary.

A simpler alternative would be to always allow symlinks at depth 1 without a setting, since the exposure is pretty limited at that level.                              

Either way I'm happy to implement it and open a PR once we agree on the direction!

BenCello:
## Summary

Fixes #259

Project directories inside `projects/` that are symlinks are now included in the scan. This supports setups where sessions are shared across multiple macOS user accounts via symlinks into a shared folder (e.g. `/Users/Shared/.claude/projects/`).

The change is a single filter in `scan_projects`:

```rust
// Before: only real directories
.filter(|e| e.file_type().is_dir())

// After: real directories + symlinks that resolve to directories
.filter(|e| {
    e.file_type().is_dir()
        || (e.file_type().is_symlink() && e.path().is_dir())
})
```

Symlinks are **only followed at depth 1** (project level) — never deeper — so traversal cannot escape the `projects/` tree. The existing behaviour for all-real-directory layouts is unchanged.

## Test plan

- [ ] Normal project directories still appear as before
- [ ] Symlinked project directories now appear in the session list
- [ ] "Show JSONL File" on a session from a symlinked project correctly reveals the file in the real folder
- [ ] Dangling symlinks (pointing to a non-existent or non-directory target) are silently skipped

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed project scanning to properly recognize symbolic links pointing to project directories.

### DIFF
--- a/src-tauri/src/commands/project.rs
+++ b/src-tauri/src/commands/project.rs
@@ -172,7 +172,12 @@ pub async fn scan_projects(claude_path: String) -> Result<Vec<ClaudeProject>, St
         .max_depth(1)
         .into_iter()
         .filter_map(std::result::Result::ok)
-        .filter(|e| e.file_type().is_dir())
+        .filter(|e| {
+            // Accept real directories and symlinks that resolve to directories.
+            // Symlinks are only followed at depth 1 (project level), never deeper,
+            // so there is no risk of traversing outside the projects/ tree.
+            e.file_type().is_dir() || (e.file_type().is_symlink() && e.path().is_dir())
+        })
     {
         let raw_project_name = entry.file_name().to_string_lossy().to_string();
         let project_path = entry.path().to_string_lossy().to_string();


jhlee0409:
Thanks for the contribution, @BenCello! The symlink support at depth 1 fixes #259 cleanly, and the safety reasoning around `max_depth(1)` is sound — symlinks can't escape the `projects/` tree.

CodeRabbit raised two optional hardening suggestions (regression tests + canonicalize-based dedup against the rare "symlink + real dir" case). Both are valid but additive — we're keeping this PR narrowly scoped to the symlink-enable change. Merging now and tracking the hardening as follow-up in #280.

BenCello:
Hi @jhlee0409!
Sorry for my late answer. Happy to contribute.
It looks like the deduplication is already addressed and fixed. I'm not sure I'm tooled to address the regression tests for symlink, but I can take the time to ask Claude about it if you'd like :-)
Best,
Benjamin

PR #278, opened 2026-04-22
fix(ui): clamp context menus to panel bounds and close on scroll

jhlee0409:
## Summary

Follow-up to #268. That PR fixed the containing-block bug by portaling both right-click context menus to `document.body`. Two UX gaps surfaced once the menus were free-floating:

1. **Menu overflows the sidebar's right edge.** The clamp was viewport-based. With a 380px sidebar in a 1400px viewport, `cursorX + menuWidth` still fit the viewport, so no clamp fired and the menu floated out over the message viewer.
2. **Menu stays pinned when the list scrolls.** Portaled + `position: fixed` means scrolling the virtualized session list did nothing. Native desktop menus (Finder, Explorer) close on scroll. This is also a correctness concern — a virtualized `SessionItem` can unmount while the portaled menu keeps stale-closure action handlers alive.

## Changes

- `src/utils/contextMenu.ts` (new) — `computeMenuPosition` pure function. Flips the menu to the opposite side of the cursor when it would overflow the boundary, then clamps into bounds with 8px padding. Boundary is a plain `{left,top,right,bottom}` object with internal viewport fallback (test-friendly, no jsdom `DOMRect` dependency).
- `src/components/ProjectTree/index.tsx` — mark both `<aside>` branches (collapsed + expanded) with `data-menu-boundary`. Single marker covers desktop sidebar and mobile Sheet because `ProjectTree` is rendered in both.
- `src/components/SessionItem/SessionItem.tsx` + `src/components/ProjectTree/hooks/useProjectTreeState.ts` — at right-click time, resolve the nearest boundary via `e.currentTarget.closest('[data-menu-boundary]')?.getBoundingClientRect() ?? null` and pass it through. No marker found → viewport fallback (current behavior).
- `src/components/SessionItem/components/SessionContextMenu.tsx` + `src/components/ProjectContextMenu.tsx` — replace inline clamp with `computeMenuPosition`. Add `document.addEventListener("scroll", onClose, { capture: true, passive: true })` and `window.addEventListener("resize", onClose)`. Both armed one animation frame after mount so a synchronous scroll burst during the click-to-open sequence can't close the menu immediately. Cleanup matches the capture flag.

## Tests

- `src/utils/contextMenu.test.ts` — 9 unit cases: fit, flip-x, flip-y, flip-both, menu-wider-than-boundary, menu-taller-than-boundary, viewport fallback, custom padding, and positivity when menu exceeds a sidebar-rooted boundary.
- `src/test/SessionContextMenu.test.tsx` + `src/test/ProjectContextMenu.test.tsx` — 4 integration cases each: scroll-close after rAF, resize-close after rAF, pre-rAF scroll ignored, cleanup symmetry on unmount.

## Test plan

- [x] `pnpm tsc --build .` — clean
- [x] `pnpm lint` — clean (pre-existing unrelated warning in `GlobalSearchModal.tsx`)
- [x] `pnpm vitest run` — 768/768 pass (prior 751 + 17 new)
- [x] Manual: right-click session near sidebar right edge → flips left, stays in panel
- [x] Manual: right-click near viewport bottom → flips up
- [x] Manual: scroll the session list with menu open → menu closes
- [x] Manual: resize window with menu open → menu closes
- [x] Manual: drag-resize sidebar with menu open → menu closes via existing outside-click
- [x] Manual: outside-click + ESC still close (regression)
- [x] Manual: every menu action still works (Rename, Copy Session ID, Delete, Copy path, Hide project, etc.)

## Summary by CodeRabbit

* **New Features**
  * Context menus now close on scroll or resize (with a short safeguard to avoid immediate closure during opening).
  * Context menus are positioned more intelligently and now respect nearby panel/container boundaries to avoid overflow.

* **Tests**
  * Added comprehensive tests covering positioning, boundary handling, flipping/clamping behavior, and the scroll/resize close workflow.

### DIFF
--- a/src/components/ProjectContextMenu.tsx
+++ b/src/components/ProjectContextMenu.tsx
@@ -6,10 +6,11 @@ import { useTranslation } from "react-i18next";
 import { toast } from "sonner";
 import { cn } from "@/lib/utils";
 import type { ClaudeProject } from "../types";
+import { computeMenuPosition, type Boundary } from "@/utils/contextMenu";
 
 interface ProjectContextMenuProps {
   project: ClaudeProject;
-  position: { x: number; y: number };
+  position: { x: number; y: number; boundary?: Boundary | null };
   onClose: () => void;
   onHide: (projectPath: string) => void;
   onUnhide: (projectPath: string) => void;
@@ -26,7 +27,7 @@ export const ProjectContextMenu: React.FC<ProjectContextMenuProps> = ({
 }) => {
   const { t } = useTranslation();
   const menuRef = useRef<HTMLDivElement>(null);
-  const [adjustedPosition, setAdjustedPosition] = useState(position);
+  const [adjustedPosition, setAdjustedPosition] = useState({ x: position.x, y: position.y });
 
   // Close on click outside
   useEffect(() => {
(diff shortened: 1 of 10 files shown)


jhlee0409:
Thanks for the review. Ran a second independent pass via `codex exec` and both verdicts line up, so addressing as follows:

**Nitpick 1 (extract `useCloseOnScrollOrResize` hook): deferring.**
Two callsites, both deliberately parallel, and the close-on-scroll pattern for portaled menus is unlikely to get tweaked. Rule-of-three applies here — the actual drift risk is low (any change to one menu's close semantics would almost certainly touch the other, since they share UX) and the coordination cost is bounded to the `{capture: true}` symmetric cleanup, which is now explicitly called out in the effect's comment. If a third context menu appears (or we migrate to Radix `ContextMenu`), that's the natural moment to extract. Codex's second opinion: *"two tightly parallel callsites do not justify a hook yet; extracting this specific effect mainly centralizes a small implementation detail rather than reducing meaningful maintenance risk."*

**Nitpick 2 (JSDoc + overflow behavior): applied in f359f8b.**
The "menu wider than boundary → pins to left/top + padding" behavior is intentional — `src/utils/contextMenu.test.ts`'s 9th case (`keeps x within a sidebar-rooted boundary when the menu is wider than the panel`) explicitly asserts `expect(result.x).toBe(sidebar.left + 8)` for exactly this scenario. The new JSDoc makes that policy explicit at the call site so future readers don't have to trace the math.

Issue #293, opened 2026-05-06
Antigravity provider: hardening follow-ups (after #291)


jhlee0409:
Tracks the path-validation / input-validation hardening items deferred from #263 / #291.

The realistic path-traversal class was closed in #291 (symlink guards everywhere; `load_messages` requires the path to resolve to the detected root). These items deepen that defense.

## Items

- [ ] **`antigravity_root_from_path` default fallback weakens validation** — When the marker `.token-monitor/rpc-cache/v1` cannot be found, the function falls back to `default_root`. That makes any path under the default root pass the `path_in_resolved` check in `load_messages` trivially, regardless of whether it's actually a valid antigravity directory. The function should return `None` on no-marker.

- [ ] **`root_path` fallback in `get_antigravity_project_summary`** — Same shape: when `antigravity_root_from_path(root_path)` returns `None`, the code accepts `PathBuf::from(root_path)` directly. Tied to the item above.

- [ ] **`session_id` input validation tightening** — `is_valid_antigravity_session_id` already enforces `[A-Za-z0-9_-]+`, non-empty. Confirm every command/file-path use site already calls it (currently only HashMap-key use is verified). Tighten if any path-component use slips in.

- [ ] **u64 → i64 token cast** — Token totals are u64 internally but cast to i64 at frontend boundaries. Add a saturating cast or guard against pathological inputs.

jhlee0409:
Additional defense-in-depth items raised by CodeRabbit on #291 (deferred from there):

- [ ] **Symlinked `monitor-state*.json` deserialization** — `load_active_state` and `load_archive_states` only check existence/name then `read_to_string` + `serde_json::from_str`. A symlinked file under the antigravity root still gets followed. Add `symlink_metadata` check before reading.
- [ ] **Symlinked `usage.jsonl` / `steps.jsonl` in RPC cache** — `parse_token_files()` admits files via `exists()` only. Even though parent dirs are symlink-guarded, the files themselves can still be symlinks pointing outside. Filter via `file_type().is_symlink()` before parsing.
- [ ] **Canonicalize `session_path` before reading `usage.jsonl`** — `marker_rooted_path()` validates by textual prefix only. A symlinked session directory inside the detected root still passes, and `dir.join("usage.jsonl")` is then read directly. `fs::canonicalize()` would close this TOCTOU/symlink-bypass class.
[...]

Issue #292, opened 2026-05-06
Antigravity provider: stats correctness follow-ups (after #291)


jhlee0409:
Tracks the stats-correctness items deferred from #263 / #291.

These don't block #291 since the provider works for the basic case and surfaces non-zero numbers; they fix specific accuracy issues.

## Items

- [ ] **Global summary missing reasoning aggregation** — `summary.token_distribution.reasoning` is not accumulated from `stats.token_distribution.reasoning`, so the UI's reasoning breakdown stays at 0 even when sessions report reasoning tokens. (`src-tauri/src/commands/stats.rs` global-summary block)

- [ ] **Antigravity tool aggregation ignores date filter** — `most_used_tools` for Antigravity reads full sessions via `load_messages()` and aggregates regardless of the active `start_date`/`end_date`, while token totals respect the filter. Dates filter applied to one but not the other.

- [ ] **Daily / activity stats ignore stats_mode** — In Antigravity per-project stats, `session_stats` honors the conversation-only / billing-total mode but daily and heatmap aggregations re-use raw `record.total_tokens` / `record.input_tokens`, making header totals diverge from per-day totals.

- [ ] **External root excluded from global stats** — `resolve_antigravity_root()` is honored elsewhere but the global stats path only walks `~/.gemini/antigravity`, so an external state directory contributes 0 tokens at the global level.

- [ ] **`groups` missing → input/cache zeroed** — When a record carries `estimatedTokensUsed` but no `tokenBreakdown.groups`, the conversation-only path returns `Some((0, total))` and zeroes input/cache. Should fall back to the full input/cache totals like the non-detail path.

- [ ] **`last_poll_at` lost in merge_states** — `merge_states` hardcodes `last_poll_at: None`, dropping the active state's polling timestamp before the frontend sees it.

jhlee0409:
CodeRabbit also flagged the following on #291 (already covered by items above but documenting source for traceability):

- `summary.token_distribution.reasoning` aggregation in the global summary loop (covered by item 1)
- `antigravity_chat_token_breakdown` returning `Some((0, total))` when `tokenBreakdown.groups` is missing — same as item 5 ("groups missing → input/cache zeroed")
- Global stats Antigravity collection using `get_antigravity_root()` directly instead of `resolve_antigravity_root()` — covered by item 4 (external root excluded), with the additional note that switching the resolver is the concrete fix

jhlee0409:
[...]

PR #318, opened 2026-05-13
fix(antigravity): rpc-cache fallback in load_antigravity_usage_records (#292 item 7)

jhlee0409:
## Summary

Resolves **item 7** of #292 — `load_antigravity_usage_records` in `commands/stats.rs` now mirrors the rpc-cache fallback already used by `providers::antigravity::load_messages`, so brain/-only sessions contribute records (and therefore tokens) to per-session / project / global stats instead of silently zeroing out.

## Why

`providers::antigravity::load_messages(session_path)` resolves the session's token data via a two-step lookup:

1. `<session_path>/usage.jsonl` (filesystem source)
2. `<root>/.token-monitor/rpc-cache/v1/<session_id>/usage.jsonl` (rpc-cache fallback)

`load_antigravity_usage_records` only ever did step 1. For a session whose `file_path` points at a brain/-only directory (no in-place `usage.jsonl`), `load_messages` would surface the conversation while the stats pipeline saw `Ok(vec![])` and treated it as 0 tokens.

## What changed

- **`providers::antigravity::resolve_usage_jsonl_path(session_path) -> Option<PathBuf>`** — `pub(crate)` helper that captures the two-step lookup. `load_messages` now delegates to it.
- **`commands::stats::load_antigravity_usage_records`** — calls the same helper instead of inlining `PathBuf::from(session_path).join("usage.jsonl")`.

No public API change. Direct-`usage.jsonl` behaviour is unchanged; only the fallback path is newly available to the stats helper.

## Tests added

- `resolve_usage_jsonl_path` (in `providers::antigravity::tests`):
  - prefers `<session_path>/usage.jsonl` when present
  - falls back to the rpc-cache `usage.jsonl` for brain/-only sessions
  - returns `None` when neither candidate exists
- `load_antigravity_usage_records` (in `commands::stats::tests`):
  - brain/-only session resolves via the rpc-cache (parses real record fields)
  - missing-everywhere still returns `Ok(vec![])`

## Scope notes

- `#292` stays open. Items 1 (reasoning aggregation), 2 (date filter on `most_used_tools`, in flight as #317), and 3 (daily/heatmap `stats_mode`) are tracked separately.
- Symlink hardening on the rpc-cache fallback (canonicalize / `symlink_metadata`) is covered by **#293 item 7** — explicitly deferred so the surface area of this PR stays small.

## Test plan

- [x] `cargo fmt` clean locally
- [ ] `cargo clippy -- -D warnings` — could not run locally (`tauri-runtime-wry-2.10.1` + local rustc compile error, independent of this diff; CI handles Rust validation)
- [ ] CI: lint + tests on PR
- [ ] Manual smoke: configure an Antigravity session whose `file_path` points at a brain/-only directory; verify it now contributes non-zero tokens to per-project / global stats.

> *This was authored by Claude during a /loop session.*

## Summary by CodeRabbit

* **Refactor**
  * Improved session usage record resolution with enhanced fallback handling to ensure records load correctly across different session configurations
  * Added unit tests covering usage record resolution including edge cases

[Review Change Stack](https://app.coderabbit.ai/change-stack/jhlee0409/claude-code-history-viewer/pull/318)

### DIFF
--- a/src-tauri/src/commands/stats.rs
+++ b/src-tauri/src/commands/stats.rs
@@ -1389,10 +1389,9 @@ fn calculate_session_active_minutes(timestamps: &mut [DateTime<Utc>]) -> u32 {
 fn load_antigravity_usage_records(
     session_path: &str,
 ) -> Result<Vec<AntigravityUsageRecord>, String> {
-    let usage_path = PathBuf::from(session_path).join("usage.jsonl");
-    if !usage_path.exists() {
+    let Some(usage_path) = providers::antigravity::resolve_usage_jsonl_path(session_path) else {
         return Ok(vec![]);
-    }
+    };
 
     let content = fs::read_to_string(&usage_path)
         .map_err(|e| format!("Failed to read {}: {}", usage_path.display(), e))?;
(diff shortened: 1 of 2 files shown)


jhlee0409:
@coderabbitai @copilot — thanks for the catch. The full hardening (re-order so `marker_rooted_path` validates the canonical dir before any IO + `symlink_metadata`-gated `admit_usage_jsonl` for both candidate locations) is now in the stacked PR #319 (commit 14c90a5).

Keeping this PR scoped to the rpc-cache fallback itself; the symlink/canonical-form defenses land in #319 so the diffs stay readable.

PR #319, opened 2026-05-13
fix(antigravity): canonicalize session_path in resolve_usage_jsonl_path (#293 item 7)

jhlee0409:
> **Stacked on #318.** Targets that branch — merge #318 first, then this PR will rebase onto `develop` automatically and become mergeable to `develop`. The diff below shows only commits unique to this branch.

## Summary

Resolves **item 7** of #293 — the canonicalisation step that closes the symlinked-session-dir bypass class in `resolve_usage_jsonl_path` (introduced by #318 for the rpc-cache fallback, but the underlying read shape predates this PR).

## Why

`marker_rooted_path` validates by walking up looking for a marker directory, or by textual prefix against the detected root. Both checks operate on the **textual** form of `session_path`. A symlinked session directory inside the antigravity root would pass either check while pointing outside the root — and the subsequent `dir.join("usage.jsonl")` would follow the symlink and read the attacker-controlled target.

Example: an attacker who can place files inside `<antigravity_root>/brain/` could `ln -s /etc <antigravity_root>/brain/session-evil`, and a request scoped to `<antigravity_root>/brain/session-evil` would end up reading `/etc/usage.jsonl` (or fall through to other reads).

## What changed

`providers::antigravity::resolve_usage_jsonl_path`:

- Canonicalises the in-session candidate before reading. The canonical form is the path with all directory symlinks dereferenced.
- Re-runs `marker_rooted_path` against the canonical form. If the canonical path no longer falls under any recognised antigravity root, return `None`.
- The rpc-cache fallback is unchanged: it builds its file path under the marker-rooted root by construction, so a textual `session_id` cannot escape — the file always lives under the trusted root regardless of what was passed in.

No public API change.

## Out of scope (deferred)

- **File-level symlink check on `usage.jsonl` itself.** `parse_token_files` already gates rpc-cache files via `symlink_metadata` (#314), but `resolve_usage_jsonl_path` / `load_messages` do not. Adding it here would extend the scope of the issue. Track as follow-up if reviewers want it.
- **Windows `\\?\` prefix interaction.** On Windows `fs::canonicalize` returns extended-length paths. `marker_rooted_path`'s walk-up does not depend on textual matching (it `.exists()` on a directory marker at each ancestor), so the canonical form should still resolve correctly. The textual-prefix fallback inside `marker_rooted_path` is reached only when no marker is found; with #316 in place that path returns `None` cleanly. No code path requires comparing a canonical path against a non-canonical default root.

## Tests

- Existing `test_resolve_usage_jsonl_path_prefers_in_session_file` now canonicalises both sides of the assertion (covers macOS `/var` → `/private/var` and any other symlinked temp-dir prefix).
- New Unix-only regression `test_resolve_usage_jsonl_path_rejects_symlink_escaping_root`: a `brain/<sid>` entry that is a symlink to an outside-of-marker-root directory containing `usage.jsonl` must resolve to `None`.
- `test_resolve_usage_jsonl_path_falls_back_to_rpc_cache` and `test_resolve_usage_jsonl_path_returns_none_when_missing_everywhere` (from #318) still pass — the rpc-cache fallback path is unchanged.

## Scope notes

- `#293` stays open. Items 1, 2, 3, 5, 6, 8, 9 already landed (#314, #315, #316). Item 4 was [verified stale](https://github.com/jhlee0409/claude-code-history-viewer/issues/293#issuecomment-4439444479). Once this PR lands, **#293 can close**.

## Test plan

- [x] `cargo fmt` clean locally
- [ ] `cargo clippy -- -D warnings` — could not run locally (`tauri-runtime-wry-2.10.1` + local rustc compile error, independent of this diff; CI handles Rust validation)
- [ ] CI: lint + tests on PR
- [ ] Manual smoke: drop a symlinked `brain/session-X` pointing at an arbitrary directory containing `usage.jsonl`; confirm stats commands now refuse to read it instead of surfacing tokens.

> *This was authored by Claude during a /loop session.*

### DIFF
--- a/src-tauri/src/providers/antigravity.rs
+++ b/src-tauri/src/providers/antigravity.rs
@@ -433,13 +433,26 @@ pub fn load_sessions(path: &str, _exclude_sidechain: bool) -> Result<Vec<ClaudeS
 /// and fall back to the rpc-cache location for filesystem-only / brain-only
 /// sessions. Returns `None` when the session path does not resolve to a
 /// recognised Antigravity root, or when neither candidate exists.
+///
+/// Canonicalises `session_path` before reading so a symlinked session
+/// directory cannot bypass [`marker_rooted_path`]'s textual root check by
+/// pointing outside the antigravity root.
 pub(crate) fn resolve_usage_jsonl_path(session_path: &str) -> Option<PathBuf> {
     let dir = PathBuf::from(session_path);
-    let usage_path = dir.join("usage.jsonl");
-    if usage_path.exists() {
-        return Some(usage_path);
+
+    // Prefer an in-session `usage.jsonl`. Canonicalising first dereferences
+    // any directory-level symlinks; a candidate that resolves outside a
+    // marker-rooted antigravity root is rejected.
+    if let Ok(canonical_dir) = std::fs::canonicalize(&dir) {
+        let usage_path = canonical_dir.join("usage.jsonl");
+        if usage_path.exists() {
+            return marker_rooted_path(&canonical_dir.to_string_lossy())
+                .map(|_| usage_path);
+        }
     }
 
+    // Fallback: rpc-cache. The file lives by construction under the
+    // marker-rooted antigravity root, so the textual session_id is safe.
     let root = marker_rooted_path(session_path)?;
     let session_id = dir.file_name()?.to_string_lossy().to_string();
     let rpc_cache = get_antigravity_rpc_cache_root(&root)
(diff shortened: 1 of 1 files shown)


jhlee0409:
Addressed review feedback in 14c90a5:

1. **Reordered root check** (Copilot's primary inline comment): `marker_rooted_path(canonical_dir)` now runs *before* any filesystem probe on `usage.jsonl`. Out-of-root candidates are rejected without touching the target path.

2. **File-level symlink check** (CodeRabbit's "Quick win" on #318 + Copilot's suppressed comments on #319): new `admit_usage_jsonl` helper gates both the in-session and rpc-cache candidates via `symlink_metadata`, refusing symlinks and non-regular files. Consistent with the pattern established by #314 for `parse_token_files`.

Tests:
- `test_resolve_usage_jsonl_path_rejects_symlinked_usage_jsonl` (Unix) — symlinked `usage.jsonl` inside a legitimate session dir
- `test_resolve_usage_jsonl_path_rejects_symlinked_usage_jsonl_in_rpc_cache` (Unix) — symlinked `usage.jsonl` in the rpc-cache fallback

Both pre-existing tests (canonical-form preference, rpc-cache fallback, directory-level symlink escape) still pass.

jhlee0409:
Superseded by #320 — same commits rebased onto current `develop` after #318's merge auto-closed this PR.

Issue #121, opened 2026-02-10
bug: fails to download the new version and error message is shown in Korean


SoraDaibu:
# Issue
downloading fails 

# Expected Behavior
downloading succeeds. even if it fails, error message should be English or the configured language. 

# Screen Video

https://github.com/user-attachments/assets/fd2bdd9a-63c4-4b73-a350-e2527547ad42

SoraDaibu:
FYI, I was able to download the new version by deleting the old version app.

PR #325, opened 2026-05-14
fix(updater): native OS-level relaunch fallback for Tauri v2 macOS bug (#287)

### LINKED ISSUE #287: [Bug Report] Updater failed (1.5.3 → 1.11.0)
I still can't update the app. the same issue remains as https://github.com/jhlee0409/claude-code-history-viewer/issues/121. 

[Updater Diagnostics]
timestamp=2026-05-05T09:13:10.235Z
error=update.manual_restart_required
currentVersion=1.5.3
newVersion=1.11.0
hasUpdate=true
isDownloading=false
isInstalling=false
isRestarting=false
downloadProgress=100
userAgent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)

---
**System Information**
- App Version: 1.5.3
- OS: macos Unknown
- Architecture: aarch64

SoraDaibu:
Closes #287 (recurrence of #121).

## The bug, in plain English

On macOS, when you tap "Download & Install" in the updater, the new app is actually copied into `/Applications` correctly — but Tauri v2's `relaunch()` then throws because of upstream bugs (tauri-apps/tauri#13923, #11392, #8472). The dialog catches the error and shows a green "Please quit and reopen" banner, but its primary button is labelled **Close** and only dismisses the dialog. People hit Close, keep using the app, and conclude the update silently failed. Diagnostic in #287: `downloadProgress=100`, `isDownloading=false`, no auto-relaunch.

## The fix, in plain English

If Tauri's relaunch path throws after a successful download, we now try a second, OS-native restart that we control end to end. If even that fails, the dialog finally surfaces a real **Restart Now** button (instead of "Close") that tries it one more time on demand. The "quit manually" message is the last resort, not the first response.

## What changed (3 topical commits)

1. **`feat(updater): add force_quit_and_relaunch Tauri command`**
   New Rust command in `src-tauri/src/commands/update.rs`. Spawns a detached shell helper that waits ~1s, then re-opens the bundle:
   - macOS: `sh -c 'sleep 1 && open -n <app>.app'`
   - Windows: `cmd /C "ping -n 2 127.0.0.1 > /dev/null && start \"\" \"<exe>\""`
   - Linux: `sh -c 'sleep 1 && <exe> &'`
   Then `AppHandle::exit(0)` 200 ms later so the helper outlives us and the single-instance lock is released cleanly. Two unit tests cover `shell_escape` (single-quote wrapping + embedded-quote escaping).

2. **`fix(updater): auto-invoke force_quit_and_relaunch when Tauri relaunch fails`**
   In `useUpdater.ts`'s catch block, when `downloadCompleted` is true we now `invoke('force_quit_and_relaunch')` _before_ falling into the manual-restart UX. Three new vitest cases: fallback succeeds, both paths fail, and partial-download (fallback must NOT run). Existing relaunch-failure tests updated to assert the fallback was attempted.

3. **`feat(updater): replace "Close" with "Restart Now" button on manual-restart fallback`**
   `SimpleUpdateModal.tsx` primary button label flips to **Restart Now** when `requiresManualRestart` is true and runs the same helper. On second failure, a toast (`common.error.updateRelaunchFailed`) explains the manual ⌘Q / Alt+F4 path and the dialog closes. Two new i18n keys translated across all five locales (en/ko/ja/zh-CN/zh-TW); `types.generated.ts` regenerated.

## Quality gate (per CLAUDE.md)

All green locally:
- `pnpm tsc --build .` ✅
- `pnpm vitest run` → **788 / 788** passing (21 in `useUpdater.test.ts`)
- `pnpm lint` ✅
- `pnpm run i18n:validate` → **1772 keys** synced across 5 locales
- `cargo test -- --test-threads=1` → **557 + 2 new** passing
- `cargo clippy --all-targets --all-features -- -D warnings` ✅
- `cargo fmt --all -- --check` ✅
- pre-commit hooks (lint-staged) passed on every commit

## Test plan for reviewers

- **Unit-level**: the three new vitest cases isolate each branch (auto-fallback, double-failure, partial download). Mock `@tauri-apps/api/core` invoke + `relaunch` to exercise.
- **Visual smoke** (no release needed): run `just dev`, temporarily make `relaunch()` always throw — confirm the dialog never appears (fallback wins). Then also make the invoke reject — confirm the **Restart Now** button appears with the rotating icon, and clicking it surfaces the toast.
- **Real end-to-end**: only possible after cutting a release strictly newer than the current latest, so it'll happen naturally on the next release tag from `develop`.

## Out of scope

- The upstream Tauri `relaunch()` bug itself (we route around it).
- The download/install flow (unchanged).
- Version bump — PR targets `develop` per the branch strategy, not `main`.

## Risk assessment

- New code path only fires when the existing flow already threw, so the happy path is byte-identical to today.
- `force_quit_and_relaunch` exits the process; if the helper somehow fails to spawn we set `requiresManualRestart=true` and the user is in exactly the state they'd be in today — no regression.
- macOS path uses `open -n` which is the same mechanism `tauri-cli` uses for `tauri dev`; well-trodden.

## Summary by CodeRabbit

* **New Features**
  * "Restart Now" button in the update modal now attempts an automatic relaunch and keeps the dialog open until restart completes; shows a relaunch-failure message if it fails.

* **Reliability**
  * Added a detached automatic relaunch fallback to improve restart reliability; if it fails the app falls back to the manual-restart flow and surfaces a clear error.

* **Localization**
  * Added restart label and relaunch-failure translations for EN, JA, KO, ZH-CN, ZH-TW.

* **Tests**
  * Expanded tests to cover fallback relaunch success and failure scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jhlee0409/claude-code-history-viewer/pull/325?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src-tauri/src/commands/mod.rs
+++ b/src-tauri/src/commands/mod.rs
@@ -11,6 +11,7 @@ pub mod session;
 pub mod settings;
 pub mod stats;
 pub mod unified_presets;
+pub mod update;
 pub mod watcher;
 pub mod wsl;
 
--- a/src-tauri/src/commands/update.rs
+++ b/src-tauri/src/commands/update.rs
@@ -1,4 +1,14 @@
+use std::process::Stdio;
 use tauri::AppHandle;
+
+#[cfg(target_os = "windows")]
+use std::os::windows::process::CommandExt;
+
+// https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags
+#[cfg(target_os = "windows")]
+const DETACHED_PROCESS: u32 = 0x0000_0008;
+#[cfg(target_os = "windows")]
+const CREATE_NEW_PROCESS_GROUP: u32 = 0x0000_0200;
 
 /// Force-quits the current app process and spawns a detached helper that
 /// re-launches the app after the current process exits.
@@ -7,9 +17,20 @@ use tauri::AppHandle;
 /// upstream bugs on macOS (tauri-apps/tauri#13923, #11392, #8472) — after a
 /// successful download and install, the relaunch step sometimes fails and the
 /// user is stranded on the old binary even though the new bundle is on disk.
+///
+/// The helper polls the parent PID and only relaunches once the parent has
+/// fully exited. This is required because `tauri-plugin-single-instance` would
+/// otherwise see the parent still running, refocus its window, and let the new
+/// process exit without applying the update.
 #[tauri::command]
 pub fn force_quit_and_relaunch(app: AppHandle) -> Result<(), String> {
     let current_exe = std::env::current_exe().map_err(|e| format!("current_exe failed: {e}"))?;
+    let ppid = std::process::id();
+
+    log::info!(
+        "[updater] force_quit_and_relaunch ppid={ppid} exe={}",
+        current_exe.display()
+    );
 
     #[cfg(target_os = "macos")]
     {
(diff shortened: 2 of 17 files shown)


SoraDaibu:
Thanks @coderabbitai — addressed in 54c6862 and ae3f0b8.

| Comment | Status | Where |
|---|---|---|
| `SimpleUpdateModal.tsx`: don't `onClose()` when Restart Now fails — banner has the only manual-recovery guidance | ✅ Fixed | ae3f0b8 |
| `update.rs`: single-instance race between helper relaunch and parent exit | ✅ Fixed — helper now polls parent PID (10s cap) instead of a blind `sleep 1` | 54c6862 |
| `update.rs`: Windows `cmd /C "start"` corrupts paths containing `%` via env-var expansion | ✅ Fixed — switched to PowerShell `Wait-Process` + `Start-Process` with single-quoted args; also sets `DETACHED_PROCESS \| CREATE_NEW_PROCESS_GROUP` so no console flashes | 54c6862 |
| `update.rs`: helper inherits parent stdio (SIGPIPE / console attach risk) | ✅ Fixed — `Stdio::null()` on all three platforms; Linux helper also uses `setsid` | 54c6862 |
| `update.rs`: no diagnostic logging | ✅ Fixed — `log::info!` / `log::error!` around each spawn path; surfaces in `tauri-plugin-log` output | 54c6862 |
| `useUpdater.ts`: defensive 10s timeout when `force_quit_and_relaunch` resolves but exit doesn't happen | ⏭️ Skipped — for this edge case to bite, `app.exit(0)` would have to fire-and-not-fire (`AppHandle::exit` is synchronous on the loop thread). Adding a `useRef`-backed clearable timer is non-trivial in this hook (no existing unmount cleanup) and would mask a deeper Tauri bug rather than fix it. Happy to revisit if you've seen this happen in practice. |

Quality gate re-run:
- `cargo clippy --all-targets --all-features -- -D warnings` ✅
- `cargo test -- --test-threads=1` ✅
- `pnpm tsc --build .` ✅
- `pnpm vitest run src/hooks/useUpdater.test.ts` → 21/21 ✅

SoraDaibu:
Hi @jhlee0409, I've fixed the issue https://github.com/jhlee0409/claude-code-history-viewer/issues/287. this change should solve it. I appreciate if you could review it. thanks for the nice project again. hope this projects grows well for you.

jhlee0409:
Thanks for the deep dive on this one. Platform-gating via `#[cfg(target_os = ...)]` was the right call, and the CodeRabbit follow-ups (PID polling, Windows `%` escape, `Stdio::null()`, `setsid`) all addressed the right concerns. Merging now — #287 should auto-close.

PR #333, opened 2026-05-15
feat: add recoverable history sources for remote, Podman, and backups

stoneproud:
## Summary

This PR adds a recoverable multi-source history ingestion layer for Claude Code History Viewer.

It expands the viewer from "scan the current machine's default provider folders" into a model that can ingest, label, de-duplicate, and display histories from:

- local default provider directories
- user-configured custom provider roots
- SSH-synced remote hosts
- Podman containers on remote/Linux environments
- restored old-machine backups
- OpenCode SQLite databases, including live or locked databases

The user-facing goal is to make AI coding histories portable across machines and environments without requiring users to manually understand every provider's storage layout.

## Contributor Context

I want to be transparent about how this PR was produced.

Most of the implementation was written with AI coding agents, with me acting as the product/user-side driver: describing the recovery workflow I needed, testing the app on my own machine, checking whether restored/remote/Podman histories appeared correctly, and iterating on bugs found during that process.

My own coding background is limited. I have some basic TypeScript knowledge, but I do not have deep experience with this project's full web/Tauri/Rust stack. Because of that, I would especially appreciate careful maintainer review of the Rust-side sync, path handling, backup/restore, and provider integration logic.

The reason I am investing in this feature is that I use this viewer personally, and I am also using it as a harness component around AI coding workflows. I recently had to move between machines and recover AI assistant histories from old backups, local Windows paths, WSL/Podman storage, and OpenCode databases. That real recovery workflow is what drove the scope of this PR.

So this PR is not a polished maintainer-authored design from deep familiarity with the codebase. It is a user-driven, AI-assisted contribution that has been manually tested against my actual history data and environments. I am happy to adjust the implementation to better match the project's architecture and review expectations.

## Why

In real usage, AI coding assistant histories are often spread across several places:

- the current machine
- an old machine backup
- WSL or Podman development environments
- remote hosts used for coding agents
- OpenCode SQLite databases that may still be in use

Before this PR, those cases were either manual, fragile, or easy to duplicate:

- users had to restore provider folders by hand
- remote or container histories were not represented as first-class sources
- the same history could appear multiple times through local, UNC, remote-cache, or restored paths
- OpenCode virtual paths such as `opencode://global/...` were copied as non-actionable paths
- locked OpenCode databases could fail to load
- provider virtual paths could trigger subagent-loading errors or leave the main message pane blank even when the navigator had messages

This PR treats "where history came from" as part of the data model, not just a path string.

### Remote Source Sync

Adds a remote-source model and SSH/SFTP sync pipeline.

Key pieces:

- `src-tauri/src/remote/source.rs`
- `src-tauri/src/remote/sftp_client.rs`
- `src-tauri/src/remote/sync.rs`
- `src-tauri/src/commands/remote_sync.rs`
- `src/components/RemoteSources/RemoteSourcesDialog.tsx`
- `src/types/core/remoteSource.ts`

Remote sources can be configured with host, port, username, auth, system kind, and custom paths. The sync engine connects over SSH/SFTP, resolves configured/default provider roots, and pulls only provider-whitelisted history files into the local app cache.

This avoids copying entire home directories and keeps the synced data focused on the files the viewer needs.

### Provider-Aware Sync Whitelist

The remote sync engine uses provider-specific file whitelists instead of a blind recursive copy.

Examples:

- Claude Code: project/history files needed by the viewer
- Codex CLI: session history files
- OpenCode: `opencode.db`, WAL/SHM sidecars, `storage/`, and `snapshot/`

This keeps remote sync smaller and reduces accidental ingestion of unrelated files.

### Podman Container History Sync

Adds Podman-aware discovery and sync support for Linux remote sources.

The sync flow can:

- detect `podman` on a remote Linux host
- list running containers
- probe provider roots inside containers
- collect a manifest with `podman exec`
- stage whitelisted files via `podman cp`
- pull the staged files back through SFTP
- register each container-backed provider root with source metadata

Container-backed projects receive source labels such as `Podman: <container> @ <user>@<host>`, so the UI can distinguish them from local and plain SSH-synced histories.

### Source Metadata Model

Extends the project/session/custom-path data model with source information.

This lets the scanner and UI distinguish:

- local default provider data
- user custom paths
- remote host cache data
- Podman container data
- restored backup data

The source metadata is used for display labels, debug labels, custom path registration, and duplicate suppression.

### Project Tree UI Improvements

Updates the left project tree to make multi-source histories usable.

Changes include:

- compact provider badges
- adaptive truncation for long project names and paths
- reusable `TruncatedScrollText`
- source labels on project rows
- a project refresh button
- refresh flow that reloads metadata, re-detects providers, and rescans projects
- expanded project search over:
  - project name
  - project path
  - provider
  - custom label
  - source label
  - loaded session summary/id/file path

These changes make remote/restored/container projects easier to identify and find.

### Remote Source UI and Refresh Plumbing

Adds a remote-source settings dialog and refresh-status event plumbing.

The frontend can request remote refreshes and receive status updates through events, allowing the project tree to react after remote sync completes.

Remote source i18n keys are added for:

- English
- Japanese
- Korean
- Simplified Chinese
- Traditional Chinese

### OpenCode Path and SQLite Handling

Improves OpenCode support in several ways.

#### Resolve Virtual Paths Before Copying

OpenCode sessions can use virtual paths such as:

```text
opencode://global/ses_xxx
```

Those are not actionable filesystem paths. This PR adds a session path resolver so "copy file path" can return the actual backing path where possible.

#### Preserve Scoped OpenCode Global Projects

Multiple sources can each contain an OpenCode `global` project. Treating them all as `opencode://global` causes unrelated sources to collapse into one project.

This PR keeps source-scoped OpenCode projects distinct using scoped identifiers such as:

```text
opencode+path://.../global
```

#### Read Live or Locked SQLite Databases

OpenCode stores history in SQLite. Live OpenCode instances can keep the DB locked, especially through WSL/Podman/UNC paths.

The provider now tries a readonly DB open first. If that fails, it copies:

- `opencode.db`
- `opencode.db-wal`
- `opencode.db-shm`

into a temporary snapshot directory and reads the snapshot instead.

### Duplicate Suppression

Adds de-duplication for injected custom/remote/restored projects.

The scanner now:

- handles UNC path separators
- accepts `.opencode` as an OpenCode custom root
- skips nested provider roots such as `.claude/.claude`, `.codex/.codex`, and `opencode/opencode`
- prefers local default projects when the same backing data appears through injected sources
- preserves separate scoped OpenCode `global` projects from different sources

This is the main guard against duplicate sessions when the same data is visible through both local paths and mounted/restored/remote paths.

### Backup and Restore

Adds a history backup/restore feature in settings.

Key pieces:

- `src-tauri/src/commands/history_backup.rs`
- `src/components/SettingsManager/components/HistoryBackupRestore.tsx`
- `src/components/SettingsManager/UnifiedSettingsManager.tsx`

Export collects supported history roots for:

- Claude Code
- Codex CLI
- OpenCode
- remote cache data

Restore does not blindly overwrite the user's active local provider directories. Instead, it:

1. copies the backup into the app data `restored-backups` area
2. reads the backup manifest
3. resolves provider roots inside the restored backup
4. registers those roots as custom paths with source metadata

The manifest path resolver rejects absolute paths and `..` escapes so restored roots cannot escape the restore directory.

### Message Viewer Robustness

Fixes several issues discovered while validating provider virtual paths.

Changes include:

- skip subagent loading for provider virtual paths containing `://`
- return an empty subagent list for virtual paths on the Rust side
- add a static message row fallback when virtual scrolling is not ready
- extract shared row content between virtualized and static rendering
- hide message render diagnostics by default
- expose diagnostics in development with `Ctrl+Alt+D`
- restyle diagnostics so they do not visually dominate the transcript

This prevents cases where the side navigator contains messages but the main transcript area renders blank or is blocked by a subagent path error.

### Provider Source Field Consistency

Adds missing `source: None` initialization for providers touched by the new source-aware project model, including Antigravity and ForgeCode paths.

This keeps the project shape consistent across providers.

## Behavior Notes

- SSH remote source configuration is needed for remote host sync.
- Local WSL/Podman histories can also be surfaced through Windows UNC custom paths when available.
- Restored OpenCode backups are registered as custom roots instead of overwriting the active OpenCode database.
- Local default projects are preferred over injected duplicates, but distinct scoped OpenCode `global` projects are preserved.

## Testing

Automated checks run locally:

- `cargo fmt --check`
- `cargo check --manifest-path src-tauri/Cargo.toml`
- `pnpm run build`
- `pnpm lint`
  - passes with one existing warning in `GlobalSearchModal.tsx`
- Rust targeted tests:
  - `history_backup`
  - `dedupe`
  - `custom_path_provider_match`

Manual validation performed:

- restored old Claude Code / Codex / OpenCode backup data
- read Claude Code history through Windows UNC paths backed by WSL/Podman storage
- read OpenCode SQLite history through a Podman/WSL-mounted path
- verified locked/live OpenCode DB snapshot fallback
- verified project refresh updates provider/project counts after custom path registration
- verified project search can find source/session-related text
- verified OpenCode copy-path no longer returns `opencode://...`
- verified provider virtual paths no longer trigger `session_path must be an absolute path`
- verified message viewer fallback prevents a blank main pane when the navigator has messages

## Reviewer Notes

This is a large PR, but the changes are connected by one data flow:

provider storage
-> source-aware scanning
-> metadata/custom path registration
-> duplicate suppression
-> session loading
-> message rendering

Splitting these pieces too aggressively would make each individual PR harder to validate because the recovery flow only works end-to-end when the source model, scanner, UI, and provider path handling agree.

Areas most worth reviewing carefully:

- remote sync whitelist and cache layout
- SSH/SFTP auth and failure handling
- Podman discovery and staging paths
- source metadata compatibility with existing providers
- duplicate suppression behavior
- scoped OpenCode `global` project handling
- OpenCode SQLite snapshot reads
- backup restore manifest path safety
- static message fallback versus virtualized rendering behavior

Follow-up PRs can build on this foundation with additional provider coverage, remote-source UI polish, richer sync progress, and more exhaustive integration tests.

## Summary by CodeRabbit

* **New Features**
  * Remote SSH sync: per-host test/sync, manager UI, headless CLI sync, credential storage, and background sync status
  * History backup & restore with export/import flow and restore integration into projects
  * Session path resolver for virtual provider paths and a truncated-scroll marquee text component

* **Improvements**
  * Project list shows source info and compact provider dots; search matches provider/source/session data; Refresh projects + Cmd/Ctrl+R triggers sync
  * Safer sync/errors with redacted messages and clearer missing-path reporting

* **Localization**
  * Added remote sources translations for multiple languages

[Review Change Stack](https://app.coderabbit.ai/change-stack/jhlee0409/claude-code-history-viewer/pull/333)


stoneproud:
Thanks for the review. I pushed several follow-up commits to address the feedback and issues found during manual testing.

Recent follow-up changes include:

- Fixed remote credential storage through the OS credential manager.
  - Passwords/passphrases are no longer persisted in user settings.
  - Remote sync/test resolves saved credentials from the OS keychain.
  - Blank `passwordRef` / `passphraseRef` values are now treated as missing instead of causing `credential_ref is required`.

- Improved the remote source edit/test flow.
  - Testing a remote source while editing now uses the current unsaved form values.
  - A successful test no longer clears the password/passphrase field while the user is still editing.
  - Stale authentication errors are cleared after a successful test.
  - Stored credentials are now explained in the form UI so an empty field after reopening means “use the saved OS credential”.

- Addressed CodeRabbit security/robustness feedback.
  - Added safer error handling/sanitization around remote sync errors.
  - Added timeout handling for remote sync/exec paths.
  - Made manifest writes atomic.
  - Tightened path validation for backup/restore and session path resolution.
  - Rebuilt Podman staging manifests more cleanly to avoid stale synced sessions.

- Polished remote source i18n.
  - Added/cleaned remote source strings across English, Simplified Chinese, Traditional Chinese, Japanese, and Korean.
  - Wrapped the remaining visible remote source auth/system labels in translation keys.

- Revalidated locally after the latest fixes.
  - `cargo fmt --check`
  - `cargo check --manifest-path C:\cchv\src-tauri\Cargo.toml`
  - `pnpm exec tsc --noEmit`
  - `pnpm lint` passes with one pre-existing warning in `GlobalSearchModal.tsx`
  - `pnpm i18n:validate`

The PR branch has been updated in place, so this should all be visible on the existing PR.
-----------------------------------------------
Summarized and generated by AI.

jhlee0409:
Thanks for the deep iteration on this @AwakingsWings — and for being upfront about the AI-assisted authorship. The 21 commits walking through CodeRabbit's findings have addressed the bulk of the concerns very well. Before this can merge, four unresolved review threads remain — happy to merge as soon as these are addressed:

**Security (blocking):**

1. `src-tauri/src/commands/remote_sync.rs:71` — credentials can still surface in the error string returned to the frontend. Please sanitize (strip passwords / private-key paths / connection URIs) before propagation.
2. `src-tauri/src/remote/sync.rs:351` — remote-derived relative paths aren't validated before being joined to a local directory. This is a path-traversal vector via an attacker-controlled remote. Please reject `..` segments, absolute paths, drive letters, and symlinks before writing.

**Functional:**

3. `src/components/MessageViewer/MessageViewer.tsx:1117` — the full-list fallback before the virtualizer is ready will regress perf on large sessions. Please keep the virtualized path as the only render path (or gate the fallback so it never fires on large lists).
4. `src/i18n/locales/zh-TW/remoteSources.json:50` — sync rule violation flagged by `pnpm run i18n:validate`. Please re-run and align keys across all 5 locales.

No need to split the scope of the PR — happy to review the whole thing as one once those land.

stoneproud:
Thanks for the clear review. I pushed follow-up commits addressing the remaining blockers and also merged the latest `develop` back into this PR branch.

What changed:

- Sanitized remote sync/test errors before returning them to the frontend.
  - Passwords, passphrases, private key paths, and URI userinfo are redacted.
  - Authentication-related error context is collapsed to a safe public message.

- Hardened remote-derived path handling in `remote/sync.rs`.
  - Rejects unsafe relative paths containing `..`, absolute paths, drive-letter style paths, empty path segments, NUL bytes, or root/prefix components.
  - Checks local write targets for symlink components before writing.
  - Added a unit test for traversal/absolute-path rejection.

- Limited the MessageViewer static fallback to small sessions only.
  - Large sessions stay on the virtualized render path, avoiding the full-list perf regression.

- Re-ran i18n generation/validation after merging upstream `develop`.
  - `pnpm run i18n:validate` passes with all 5 locales synchronized.

- Merged latest upstream `develop`.
  - Resolved conflicts in `useSessionEditing.ts`, `useSessionEditing.test.tsx`, and `types.generated.ts`.
  - Preserved both the OpenCode virtual-path resolution behavior and the upstream resume-command CWD behavior.

- Fixed Windows local dev startup.
  - `justfile` now uses `vite.CMD` for `vite-dev` on Windows so `pnpm.CMD tauri:dev` works without requiring users to relax PowerShell execution policy.

Local validation performed:

- `cargo fmt --manifest-path src-tauri\Cargo.toml -- --check`
- `cargo check --manifest-path src-tauri\Cargo.toml`
- `pnpm.CMD run build`
- `pnpm.CMD run i18n:validate`
- `pnpm.CMD lint`
- `vitest run src/test/useSessionEditing.test.tsx`

`pnpm.CMD lint` still reports the existing `GlobalSearchModal.tsx` hook dependency warning, which was pre-existing and unrelated to this PR.
-------------------------------------------------------------------------------------------------
Summarized and generated by AI.

jhlee0409:
Thanks @stoneproud — this is a lot of genuinely high-quality work (injection-safe Podman handling, RAII tempdirs, traversal-guarded restore, 33 tests). It's not slop at all. But at 6851/-151 across 64 files spanning SSH/SFTP, Podman, SQLite, backup/restore, and UI, it's too large and too security-sensitive to merge as one unit. Two things need to happen before it can move forward:

**Required — WebUI parity (the #340/#355 regression class)**

`src-tauri/src/lib.rs` registers 5 new frontend-callable commands in `generate_handler!` (the remote sync / credential commands), but there are **no matching Axum routes** in `src-tauri/src/server/mod.rs`, and the UI isn't gated. So in `--serve`/WebUI builds these all 404. The same applies to:
- `HistoryBackupRestore` — `UnifiedSettingsManager.tsx:330` renders it with no `isTauri()` gate, and it calls `@tauri-apps/*` directly (breaks in the browser).
- `useAppKeyboard.ts:183-188` — Cmd/Ctrl+R is unconditionally hijacked (`preventDefault` + `syncRemotesAndRefresh`) with no `isTauri()` guard, overriding normal reload in WebUI.

Please either add the Axum routes **or** gate all of this behind `isTauri()`.

**Required — security decision**

SSH host-key verification appears disabled (`sftp_client.rs` accept-all) with password auth. Before we ship network sync, we need a host-key policy (TOFU / `known_hosts`). I'd like to make that call explicitly — happy to discuss the approach here.

**Required — i18n**

~9 `settings.historyBackupRestore.*` keys (and a session key) are used in TSX but missing from the locale files. They need to exist in all 5 languages.

**Request — split this up**

Could you break this into independent PRs? Roughly: (1) remote/SSH sync + credentials, (2) Podman source, (3) backup/restore. Each is reviewable on its own, and the SSH one in particular deserves a focused security review. The Podman and backup/restore pieces could likely land much sooner that way.

Really appreciate the effort here — I want to get this in, just safely and in digestible pieces.

jhlee0409:
On the SSH host-key policy I flagged above — here's the decision.

Thanks for this; the SFTP plumbing looks solid, but we can't ship `AcceptAllHandler` returning `Ok(true)`: it accepts any host key with no verification, which is a silent MITM hole on any untrusted network (public wifi, a compromised router, etc.). Owning the remote box doesn't help here — host-key checking defends the *network path* between the two machines, not the endpoints. Concretely, in password mode an on-path attacker who presents their own host key completes the handshake and then receives your password in cleartext via `authenticate_password` (it's sent over the channel you just established *with the attacker*); and because the same handler backs `exec_command` and the SFTP pull, the attacker can also feed fabricated history files straight into the local scanner/UI.

**Please switch to Trust-On-First-Use.** In `check_server_key`, compute the server key's SHA256 fingerprint (russh-keys exposes a fingerprint method on the key type — use whichever the 0.46 API provides). If we already have a stored fingerprint for this source, compare and return `Ok(stored == fp)` (hard-refuse on mismatch); otherwise record it and return `Ok(true)`. Persist the pinned fingerprint in the existing keyring store (`src-tauri/src/commands/remote_credentials.rs`) under a new `kind` such as `"hostkey_fp"` via `credential_key(source_id, kind)` — no new storage layer needed. The handler is currently a unit struct; it'll need `host`/`port`/`pinned_fp` plus a way to surface the *observed* fingerprint back out after `client::connect` (the handler is moved into the session, so a shared `Arc<Mutex<Option<String>>>` field is the easy way to read it post-connect and persist it).

**Pragmatic scope:** silent TOFU (pin the first key, hard-fail on change) is enough for this PR and needs no UI — it closes the recurring-MITM gap entirely. The nicer OpenSSH-style flow (show the `SHA256:…` fingerprint on first connect, plus a loud "host identification changed — possible MITM" dialog with an explicit replace action on mismatch) can be a follow-up; please file it as an issue and leave a `// TODO` at the mismatch branch so we don't lose it. Also please update the file header comment (the "No known_hosts checking…" note at the top of `sftp_client.rs`) since we'll now be verifying at connection time. And when you add the eventual `trust_remote_host` / replace-key command, register it in both the Tauri invoke handler (`lib.rs`) and the Axum WebUI router (`server/mod.rs`) so the WebUI build keeps parity.

stoneproud:
@jhlee0409  thx for ur suggestions about safety. and i noticed that maybe i should divied them into some independend submits firstly.This is a technical debt that deserves to be addressed as a priority.
however, for my team requirements,i have made a lot of other modification in my fork.(cause this pr is opening, i have to create & copy to fork) https://github.com/stoneproud/AgentLogBook
so i wanna do this in the next future at my fork,and have no choice but should using issue channel to provide the splited prs(right?i'm not sure)
anyway, i closed this pr now,because it will be split at future.

PR #352, opened 2026-05-28
perf(search): move FlexSearch indexing to Web Worker

Rassyan:
## Summary
  - Fixed: CPU pegged at 100% when loading/searching sessions with 47k+ messages
  - Root cause: FlexSearch `tokenize:"full"` index build on main thread via chunked setTimeout
  - Fix: moved indexing to a Web Worker; first search uses linear fallback; Worker index used once ready

  ## Changes
  - New `searchWorker.ts` — dedicated Worker for FlexSearch build + search
  - `searchIndex.ts` — Worker communication layer + linear search fallback
  - `searchSlice.ts` — async search with fallback strategy
  - `messageSlice.ts` — removed eager index build on session load
  - New `searchIndex.test.ts` — unit tests for linear search

  ## Test plan
  - [x] Open 47k message session — no CPU spike on load
  - [x] Search works immediately (linear fallback)
  - [x] After Worker completes, subsequent searches use index
  - [x] `vitest run` — 11 tests pass
  - [x] ESLint passes

## Summary by CodeRabbit

* **Improvements**
  * Search indexing now builds lazily and uses a background worker when available to reduce startup/UI blocking.
  * Immediate fallback via a fast linear scan provides instant results while background indexing builds.
  * Forward-prefix matching and refined match counting improve relevance and per-message match accuracy.
  * Search operations are now readiness-aware with guards to prevent stale results and surface errors as user notifications.

* **Tests**
  * Expanded tests for linear search behavior, indexing readiness, and clearing.

---

## Update (2026-05-31, force-pushed `b3a2f30 → 5f3fea5`)

CodeRabbit's review on this branch surfaced six findings (three inline, three outside-diff). Two rounds of fixes have been folded into the existing commit chain; this is a description-only update reflecting the latest force-push.

**Round 1 (`8788ba8`, addresses three inline findings on `searchIndex.ts` / `searchWorker.ts`):**

1. `searchMessagesAsync` had no failure path. If `search-result` never came back (worker crash, dropped postMessage), the promise stayed pending forever and `pendingSearchCallbacks` leaked. Added `worker.onerror` to resolve all pending searches with `[]`, plus a 5s per-search safety timeout. Centralized cleanup in `resolveAndCleanupSearch` so the timer and callback are always released together.
2. Index build was repeatedly restarted on consecutive keystrokes — every search input went through the `else` branch in `searchSlice` and triggered another `postMessage(build)`, wasting ~3s per duplicate build. Added a `workerBuildInFlight` flag plus `lastBuildMessagesRef` so the same messages reference is a no-op while a build is in flight or already ready.
3. `searchWorker.ts` and `searchIndex.ts` used `type X = any` with `eslint-disable` comments. Replaced with `FlexSearch.Document<SearchDoc>` typing, using `as unknown as` casts only where FlexSearch's enriched search result types don't directly match the consumed shape (per CLAUDE.md guidance on `any` replacement).

**Round 2 (`5f3fea5`, replaces `b3a2f30` — addresses two outside-diff findings on `searchSlice.ts` plus a readability follow-up):**

4. **Race condition** in `setSessionSearchQuery`: with the index still building, earlier searches go through the sync linear path and set state immediately, while later searches go through the async Worker path. Their late resolve would overwrite the more recent linear result. Added a monotonic request token (`sessionSearchRequestId`) captured at call entry; after each `await`, drop the result if the token no longer matches the latest request. The same guard runs on the catch path so a stale failure can't wipe a newer successful search.
5. **User feedback on failure**: the catch block previously only called `console.error`, leaving the user with no indication that search had failed. Per CLAUDE.md §"에러 처리", user-visible feedback (toast/alert) is required on async errors. Imported `sonner` toast and emit `toast.error` with the error message.
6. **Readability follow-up**: lifted the `requestId !== sessionSearchRequestId` check into a small `isStillLatest()` predicate, and documented on the early-return that we are intentionally NOT clearing `isSearching` here — the newer in-flight request owns that state and will reset it on its own success / failure path. Without that comment the bare `return` reads like a forgotten state reset rather than a deliberate handoff (CLAUDE.md §"예측 가능성" / §"가독성").

**Verified:** `searchIndex.test.ts` 11/11 pass; `pnpm lint` and `tsc` match baseline (no new errors or warnings introduced).

### DIFF
--- a/src/store/slices/messageSlice.ts
+++ b/src/store/slices/messageSlice.ts
@@ -17,7 +17,7 @@ import type {
 } from "../../types";
 import { AppErrorType } from "../../types";
 import type { StateCreator } from "zustand";
-import { buildSearchIndex, clearSearchIndex } from "../../utils/searchIndex";
+import { clearSearchIndex } from "../../utils/searchIndex";
 import type { FullAppStore } from "./types";
 import {
   fetchSessionTokenStats,
@@ -260,17 +260,8 @@ export const createMessageSlice: StateCreator<
       // progress 메시지를 통한 parentToolUseID ↔ subagent 매핑이 성립.
       void get().loadSubagents(sessionPath, allMessages);
 
-      // Build FlexSearch index asynchronously after UI renders
-      // The buildSearchIndex now internally uses chunked async processing
-      if ("requestIdleCallback" in window) {
-        (window as Window & { requestIdleCallback: (cb: () => void) => void }).requestIdleCallback(() => {
-          buildSearchIndex(filteredMessages);
-        });
-      } else {
-        setTimeout(() => {
-          buildSearchIndex(filteredMessages);
-        }, 0);
-      }
+      // Search index is built lazily on first search to avoid blocking UI
+      // when loading large sessions (47k+ messages with tokenize:"full" is expensive).
     } catch (error) {
       // Stale error guard: await 중 다른 세션으로 이동했으면 abandoned request의
       // 에러·로딩 상태를 현재 UI에 덮어쓰지 않음 (success path의 L212 guard 미러링)
--- a/src/store/slices/searchSlice.ts
+++ b/src/store/slices/searchSlice.ts
@@ -5,6 +5,7 @@
  */
 
 import { api } from "@/services/api";
+import { toast } from "sonner";
 import type { ClaudeMessage, SearchFilters } from "../../types";
 import { AppErrorType } from "../../types";
 import type { StateCreator } from "zustand";
@@ -73,7 +74,15 @@ export const createSearchSlice: StateCreator<
   [],
   [],
   SearchSlice
-> = (set, get) => ({
+> = (set, get) => {
+  // Monotonic request token for setSessionSearchQuery: protects against
+  // stale async results overwriting newer ones when the user types quickly
+  // and the Worker is still building (linear fallback is sync, Worker path
+  // is async - earlier async result can otherwise land on top of a later
+  // sync result).
+  let sessionSearchRequestId = 0;
+
+  return {
   ...initialSearchState,
 
   // Global search
@@ -119,6 +128,7 @@ export const createSearchSlice: StateCreator<
 
   // Session search (KakaoTalk-style navigation)
   setSessionSearchQuery: async (query: string) => {
+    const requestId = ++sessionSearchRequestId;
     const { messages, sessionSearch } = get();
     const { filterType } = sessionSearch;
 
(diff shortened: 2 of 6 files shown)


jhlee0409:
Thanks @Rassyan — moving FlexSearch indexing into a Web Worker is a solid direction, and the defensive guards around the worker lifecycle are good. Two things block merge right now:

**Required — the branch doesn't compile**

`tsc --build` fails on this branch (3 errors, all in PR files; `develop` is clean). The only check here is CodeRabbit, which doesn't run `tsc`, so the green status is misleading.

1. `src/utils/searchIndex.ts:559-560` — `(m as Record<string, unknown>)` triggers TS2352 because `ClaudeMessage` is a union. Use `as unknown as Record<string, unknown>` (this is the documented pattern in CLAUDE.md for exactly this case).
2. `src/utils/searchWorker.ts:63` — `_OutgoingMessage` is declared but unused; the `eslint-disable` silences ESLint but not `tsc`'s `noUnusedLocals` (TS6196). Reference it or remove it.

**Please confirm — behavior changes not called out in the description**

3. **Tokenize `full` → `forward`** changes search recall: the linear/substring fallback and the FlexSearch prefix index can now return different results for the same query. Intentional? If so, a note in the description would help.
4. **Match order reversed** — navigation now lands newest→oldest first, against the explicit bottom-up comment in the original code. Intentional?

**Notes (non-blocking)**

- `clearSearchIndex()` (`searchIndex.ts:609`) calls `getWorker()`, which eagerly spins up the worker on every `selectSession` even if the user never searches — slightly at odds with the "lazy indexing" goal. Guarding the clear behind an `if (worker)` check on the module-level var would keep it lazy.
- The worker / forward-tokenize path has no test coverage (the worker is `null` under vitest, so all 11 tests hit the linear fallback) — so the recall change in (3) is currently untested.

Happy to re-review as soon as `tsc` passes. Thanks!

Rassyan:
Thanks for the thorough review @jhlee0409!

All three tsc errors are fixed. Quick notes on the two behavior questions:

**forward vs full tokenize** — intentional. `full` was O(n²) and pegged CPU at 100% for large sessions. `forward` is O(n) and keeps index construction fast. The linear fallback provides full substring recall during the brief pre-index window, so the user is never stuck without results. The recall gap (e.g. "bug" not matching "debugging") is documented in the code as a conscious tradeoff for chat-history search.

**Sort order** — it was the worker that diverged (ascending). Now all three paths (main-thread FlexSearch, worker, linear fallback) sort newest-first consistently.

Also addressed the non-blocking suggestions:
- `clearSearchIndex` now guards behind `if (worker)` instead of calling `getWorker()`, keeping the lazy-init behavior
- Added tests verifying newest-first sort and mid-word substring recall

`tsc --build` is clean and all 807 tests pass locally.

jhlee0409:
Thanks @Rassyan! One thing to confirm before merge: moving FlexSearch indexing into a worker means the worker's text extractor must cover the **same** content types as the main-thread path. At a glance, several structured content types (thinking, tool_use / tool_result, and the 2025 beta content types) may not be extracted in the worker, which would silently regress search recall for those messages. Could you confirm the worker extracts the same set as the main path — ideally sharing a single extractor so they can't drift? The perf approach itself looks good.

jhlee0409:
Thank you for this contribution! 🙏 `develop` had moved on since this branch was cut, so I rebased your work onto the current `develop`, resolved the conflicts (also ported the full content-type coverage into the worker so in-session search keeps matching all types), and merged it as #401 (your authorship is preserved via `Co-Authored-By`). Closing this in favor of #401. Thanks again for improving the project! 🚀

PR #353, opened 2026-05-28
feat(provider): add CodeBuddy Code support

Rassyan:
## Summary

- Add a new provider to support [CodeBuddy Code](https://cnb.cool/codebuddy/codebuddy-code) (a Claude Code fork popular in China) conversation history
- CodeBuddy stores data in `~/.codebuddy/projects/` with a compatible directory layout but a different JSONL message format
- The provider handles all format differences transparently, converting to standard `ClaudeMessage` output

## Motivation

CodeBuddy Code (by Tencent) is a widely-used Claude Code fork in China. Its conversation data shares the same directory structure as Claude (`~/.codebuddy/projects/<project-name>/*.jsonl`) but uses a different JSONL schema:

| Field | Claude | CodeBuddy |
|-------|--------|-----------|
| `timestamp` | ISO 8601 string | Unix milliseconds (number) |
| `type` | `"user"`, `"assistant"`, `"system"` | `"message"` + `role` field |
| content item type | `"text"` | `"input_text"` / `"output_text"` |
| IDs | `uuid` / `parentUuid` | `id` / `parentId` |
| tool calls | `toolUse` / `toolUseResult` | `"function_call"` / `"function_call_result"` types |

Previously, users could add `~/.codebuddy` as a custom Claude path, but sessions would show as empty because the JSONL parser expected string timestamps and failed silently on numeric values.

## Changes

**Backend (Rust):**
- New file: `src-tauri/src/providers/codebuddy.rs` — full provider implementation (detect, scan, load sessions, load messages, search)
- `src-tauri/src/providers/mod.rs` — register `Codebuddy` variant in `ProviderId` enum
- `src-tauri/src/commands/multi_provider.rs` — add codebuddy to scan/load/search dispatch
- `src-tauri/src/lib.rs` — register `~/.codebuddy/projects` in file watcher

**Frontend (TypeScript):**
- `src/types/core/session.ts` — add `"codebuddy"` to `ProviderId` type
- `src/utils/providers.ts` — add provider label, badge style, capabilities
- `src/components/ProjectTree/index.tsx` — add to provider count map
- `src/components/ProjectTree/components/ProjectItem.tsx` — add badge color

## How it works

The provider follows the same pattern as the existing Codex provider:
1. **Detection**: checks if `~/.codebuddy/projects/` exists
2. **Scanning**: walks project directories, counts `.jsonl` files
3. **Session loading**: parses JSONL headers to extract metadata (converts numeric timestamps to RFC 3339)
4. **Message loading**: converts CodeBuddy format to standard `ClaudeMessage` (`input_text`→`text`, `function_call`→`tool_use`, etc.)
5. **Search**: full-text search across all CodeBuddy sessions

## Test plan

- [x] `cargo fmt --check` passes
- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes (0 errors)
- [x] `tsc --noEmit` passes (no frontend type errors)
- [x] Manual testing: CodeBuddy projects appear in sidebar with correct session counts
- [x] Manual testing: sessions list displays correctly when expanded
- [x] Manual testing: messages render correctly (user text, assistant text, tool calls)
- [x] CI passes

## Summary by CodeRabbit

* **New Features**
  * Added CodeBuddy provider: discover, browse, scan projects, load sessions/messages, and search.

* **Stats & Sync**
  * Included CodeBuddy in global stats and real-time file-watching for live session updates.

* **UI**
  * Added CodeBuddy provider badge and included provider in project counts/filters; improved project resolution when opening sessions.

* **Localization**
  * Added CodeBuddy translations for multiple locales.

* **Tests**
  * Updated provider tests to include CodeBuddy identifier.

* **Bug Fixes**
  * Improved session/path loading safety and handling of symlinked data.

---

## Update (2026-05-31, force-pushed `66e3416 → 6a81f25`)

CodeRabbit's second-pass review on this branch surfaced three issues; while addressing them I also tightened the regression tests so they actually exercise the production code paths. All five items are folded into a single commit (`6a81f25`); the original `feat(provider)` commit (`ee779c3`) is unchanged.

**CodeRabbit round-2 findings (all fixed):**

1. `stats.rs::detect_*_provider` used `path.contains(".codebuddy")`, which misclassifies any path that *contains* the substring (e.g. `/work/foo.codebuddy-test/...`). Replaced with a new `is_codebuddy_path` helper that anchors on `~/.codebuddy/projects` via `Path::starts_with`, matching the existing `is_antigravity_path` style.
2. `scan_projects()` walked any `*.jsonl` entry, including symlinks. The existing `load_sessions()` path already blocked symlinks; mirror the same `symlink_metadata().file_type().is_symlink()` skip in `scan_projects()` so file counts and mtimes can't leak from outside the project root.
3. `convert_function_call` wrote the tool-use entry to **both** `msg.content[]` and the top-level `msg.tool_use` field. Native Claude JSONL only carries it in `content[]`, and `stats.rs::track_tool_usage` scans both locations independently — so the duplicate doubled CodeBuddy tool counts and pushed success rate to 50%. Dropped the top-level write; `load.rs` already has a fallback that extracts from `content` when `tool_use` is `None`, so search/edits/load consumers keep working unchanged.

**Test hardening (CLAUDE.md §"높은 응집도" / §"예측 가능성"):**

4. `scan_projects_skips_symlinked_jsonl` originally hand-rolled a copy of the production walkdir + symlink filter loop in the test body — so it would have stayed green even if the real `scan_projects` regressed. Extracted `scan_projects_in(base: &Path)` from `scan_projects()` and rewrote the test to drive `scan_projects_in` with a tempdir, so the assertion now locks in the real code path.
5. `is_codebuddy_path_accepts_real_layout` did `dirs::home_dir()` and silently `return`ed (claiming success) when `$HOME` was unavailable; `is_codebuddy_path_rejects_substring_lookalikes` accidentally passed via the same null-home fallback. Extracted `is_codebuddy_path_under(path, home)` and rewrote both tests to inject a fixed `/test-home/user`, so the assertions are meaningful regardless of the runner's environment.

**New regression tests:**
- `is_codebuddy_path_rejects_substring_lookalikes` (now uses injected home)
- `is_codebuddy_path_accepts_real_layout` (now uses injected home)
- `function_call_does_not_double_write_tool_use`
- `scan_projects_skips_symlinked_jsonl` (now drives `scan_projects_in`)

**Verified:** `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test -- --test-threads=1` (572 unit + 4 integration) all pass.

---

## Update (2026-05-31, force-pushed `6a81f25 → 0949a3a`)

Folded one more issue surfaced by hands-on testing of the new build into the feature commit. The two existing commits now read as:

- `425ac98 feat(provider): add CodeBuddy Code support` — the original feature plus the sidebar handoff fix described below.
- `0949a3a fix(codebuddy): address CodeRabbit round-2 review` — unchanged from the previous round-2 amend.

**Sidebar handoff fix (folded into the feature commit, `App.tsx`):**

Selecting any session in a CodeBuddy project from the sidebar would silently no-op the project switch in `handleSessionSelect`. Root cause: the comparison `project.name === session.project_name` cannot match for CodeBuddy because the sidebar's `project.name` uses `dir_name.rsplit('-').next()` (e.g. `viewer`) while the session loader keeps the encoded form (`Users-rassyan-...-viewer`). The two never line up, so the existing equality fell through and `selectProject` was never called — the user landed on the new session with the previous project's context still active.

Replaced the equality with a `session.file_path.startsWith(project.path)` prefix match (provider-agnostic — works for every provider that lays sessions out under a project directory) and kept the name equality as a fallback for sessions whose `file_path` is unavailable. Also gated `selectProject` on `currentProject?.path !== targetProject.path` so we don't redundantly reload when already on the right project.

This belongs in the feature commit because it's the same naming-mismatch bug the loader's `project_name` already had — fixing the loader without fixing the consumer left the feature half-broken.

**Verified:** `cargo test --lib -- --test-threads=1` (572/572), `pnpm test --run` (794/794), `pnpm lint` baseline unchanged.

---

## Update (2026-06-01, force-pushed `0949a3a → 90f4ee5`)

CodeRabbit's round-3 review (against `0949a3a`) raised two findings on the previously folded sidebar-handoff change — both real, both addressed and folded back into the feature commit:

- `f1bc6f1 feat(provider): add CodeBuddy Code support` — the feature, now with the boundary-aware path matching and the user-visible toast on failure.
- `90f4ee5 fix(codebuddy): address CodeRabbit round-2 review` — unchanged.

**Path-segment boundary in `findProjectForSession` (App.tsx):**

The previous `s.file_path.startsWith(p.path)` predicate would mis-match sibling projects sharing a parent directory: a session under `/a/proj2/...` would be wrongly attributed to a project at `/a/proj`. Replaced with a boundary-aware check — after the prefix matches, the next character must be `/`, `\`, or end-of-string. Handles both POSIX and Windows separators (matches the repo's existing `/[\\\/]/` regex split convention).

**`toast.error` on session-select failure (App.tsx, `session.selectError` key):**

The `handleSessionSelect` catch path previously only called `console.error`, which violates the project guideline that every async failure needs user-visible feedback. Added `toast.error` with a new `session.selectError` key, translated across all 5 locales (en/ja/ko/zh-CN/zh-TW). The console log stays for debugging.

**Verified:** `cargo test --lib -- --test-threads=1` (572/572), `pnpm test --run` (794/794), `pnpm lint` baseline unchanged, `pnpm run i18n:validate` (1786/1786 keys synced).

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 **The unified history viewer for AI coding assistants.**
 
-Browse, search, and analyze conversations from **Claude Code**, **Gemini CLI**, **Antigravity**, **Codex CLI**, **Cline**, **Cursor**, **Aider**, **OpenCode**, and **ForgeCode** — as a desktop app or headless server. 100% offline.
+Browse, search, and analyze conversations from **Claude Code**, **Gemini CLI**, **Antigravity**, **Codex CLI**, **Cline**, **Cursor**, **Aider**, **OpenCode**, **ForgeCode**, and **CodeBuddy Code** — as a desktop app or headless server. 100% offline.
 
 [![Version](https://img.shields.io/github/v/release/jhlee0409/claude-code-history-viewer?label=Version&color=blue)](https://github.com/jhlee0409/claude-code-history-viewer/releases)
 [![Stars](https://img.shields.io/github/stars/jhlee0409/claude-code-history-viewer?style=flat&color=yellow)](https://github.com/jhlee0409/claude-code-history-viewer/stargazers)
@@ -63,7 +63,7 @@ See [Server Mode](#server-mode-webui) for Docker, VPS, and systemd setup.
 
 AI coding assistants generate thousands of conversation messages, but none of them provide a way to look back at your history across tools. CCHV solves this.
 
-**Nine assistants. One viewer.** Switch between Claude Code, Gemini CLI, Antigravity, Codex CLI, Cline, Cursor, Aider, OpenCode, and ForgeCode sessions seamlessly — compare token usage, search across providers, and analyze your workflow in a single interface.
+**Ten assistants. One viewer.** Switch between Claude Code, Gemini CLI, Antigravity, Codex CLI, Cline, Cursor, Aider, OpenCode, ForgeCode, and CodeBuddy Code sessions seamlessly — compare token usage, search across providers, and analyze your workflow in a single interface.
 
 | Provider | Data Location | What You Get |
 |----------|--------------|--------------|
@@ -76,6 +76,7 @@ AI coding assistants generate thousands of conversation messages, but none of th
 | **Aider** | Project directories | Chat history and edit logs |
 | **OpenCode** | `~/.local/share/opencode/` | Conversation sessions and tool results |
 | **ForgeCode** | `~/.forge/.forge.db` | Conversation history from SQLite database |
+| **CodeBuddy Code** | `~/.codebuddy/projects/` | Conversation history with tool calls (Claude Code fork format) |
 
 No vendor lock-in. No cloud dependency. Your local conversation files, beautifully rendered.
 
@@ -101,7 +102,7 @@ Antigravity note: the viewer resolves the Antigravity root as `~/.gemini/antigra
 
 | Feature | Description |
 |---------|-------------|
-| **Multi-Provider Support** | Unified viewer for **Claude Code**, **Gemini CLI**, **Antigravity**, **Codex CLI**, **Cline**, **Cursor**, **Aider**, **OpenCode**, and **ForgeCode** — filter by provider, compare across tools |
+| **Multi-Provider Support** | Unified viewer for **Claude Code**, **Gemini CLI**, **Antigravity**, **Codex CLI**, **Cline**, **Cursor**, **Aider**, **OpenCode**, **ForgeCode**, and **CodeBuddy Code** — filter by provider, compare across tools |
 | **Conversation Browser** | Navigate conversations by project/session with worktree grouping |
 | **Global Search** | Search across all conversations from all providers instantly |
 | **Analytics Dashboard** | Dual-mode token stats (billing vs conversation), cost breakdown, and provider distribution charts |
@@ -334,7 +335,7 @@ GET /health
 ## Usage
 
 1. Launch the app
-2. It automatically scans for conversation data from all supported providers (Claude Code, Gemini CLI, Codex CLI, Cline, Cursor, Aider, OpenCode, ForgeCode)
+2. It automatically scans for conversation data from all supported providers (Claude Code, Gemini CLI, Codex CLI, Cline, Cursor, Aider, OpenCode, ForgeCode, CodeBuddy Code)
 3. Browse projects in the left sidebar — filter by provider using the tab bar
 4. Click a session to view messages
 5. Use tabs to switch between Messages, Analytics, Token Stats, Recent Edits, and Session Board
--- a/src-tauri/src/commands/archive.rs
+++ b/src-tauri/src/commands/archive.rs
@@ -622,7 +622,7 @@ pub async fn list_archives() -> Result<ArchiveManifest, String> {
 /// * `name` - Human-readable name for the archive
 /// * `description` - Optional description
 /// * `session_file_paths` - Absolute paths to the JSONL session files to archive
-/// * `source_provider` - Provider identifier (e.g. "claude", "codex", "opencode")
+/// * `source_provider` - Provider identifier (e.g. "claude", "codex", "opencode", "codebuddy")
 /// * `source_project_path` - Filesystem path of the originating project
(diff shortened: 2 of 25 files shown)


jhlee0409:
Thanks @Rassyan — I approved and ran the full Rust CI on this PR. Results:

- ✅ **Test Suite** — passing
- ❌ **Check formatting** (`cargo fmt --all -- --check`) — fails on `src-tauri/src/providers/codebuddy.rs`
- ⏭️ **Clippy** — skipped (the job stops at the first failing step, so clippy hasn't actually been verified yet)

`rustfmt` wants three spots re-wrapped in `codebuddy.rs`:

- **L187** — the `for line in reader.lines().take(...).map_while(...)` chain should break across lines
- **L1153** — the `let project_dir = projects_root.join("Users-rassyan-…")` assignment
- **L1193** — the `let canonical_tmp = std::fs::canonicalize(...).expect(...)` line

All three are auto-fixable:

```
cd src-tauri && cargo fmt --all
```

Since clippy never got to run, please also confirm it's clean locally before pushing (CI treats warnings as errors via `RUSTFLAGS: -D warnings`):

```
cargo clippy --all-targets --all-features -- -D warnings
cargo test -- --test-threads=1
```

Push the fmt fix and I'll re-approve the CI run (fork PRs need maintainer approval each push). Everything else looks merge-ready — this formatting gate is the last thing standing between it and merge. Thanks for the solid work on this provider!

Rassyan:
Thanks @jhlee0409! Pushed the following:

- `cargo fmt --all` applied — the 3 spots in `codebuddy.rs` are re-wrapped
- `cargo clippy --all-targets --all-features -- -D warnings` confirmed clean locally
- Also refactored `ProjectItem.tsx` to use the centralized `getProviderBadgeStyle()` from `providers.ts` (same pattern as `GlobalSearchModal`, removes duplicated class conditions)

All Rust and frontend tests pass locally. Ready for CI.

Issue #365, opened 2026-06-08
Infinite page reloading on the webUI version


albertored:
#294 has been closed as fixed in version 1.13.0 but I'm still getting the same exact issue.

The 405 response on `get_session_subagents` has been indeed solved but the page keeps refreshing so that was not the cause, at least not the only one.

jhlee0409:
Fixed by #367 (merged to `develop`). The root cause was the WebUI file watcher emitting a refresh on **every** filesystem event with no change filter; #367 now compares a content signature (length + mtime) before notifying and the frontend debounces/gates updates near the bottom of the list. This also resolves the duplicates #361 and #358. The fix will ship in the next release — thanks for the report @albertored!

PR #367, opened 2026-06-13
fix(webui): stop watcher refresh loop

bugsyyu:
## Summary

Addresses #365.

`notify_debouncer_mini` collapses raw filesystem events into ambiguous `Any` events. In WebUI mode, reading an already-open session can surface as a filesystem access event on some platforms, which the server previously forwarded as `session-file-changed`. The frontend then reloaded the session, read the file again, and could enter a refresh loop.

This PR adds a watcher-side content signature cache based on file length and modified time. Watch roots are primed when the desktop/WebUI watcher starts, unchanged file events are ignored, and real content changes still update the cache and emit refresh events.

## Validation

- `corepack pnpm build`
- `cargo fmt --manifest-path src-tauri/Cargo.toml -- --check`
- `cargo test --manifest-path src-tauri/Cargo.toml --features webui-server commands::watcher::tests`
- `cargo build --manifest-path src-tauri/Cargo.toml --features webui-server`
- pre-commit `cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets --all-features -- -D warnings`
- WebUI smoke test with a temporary HOME: `/health` OK, `POST /api/load_provider_messages` returned the fixture messages, `POST /api/get_session_subagents` returned `[]`, and an active `/api/events` SSE connection received no event from the read-only message load.

## Summary by CodeRabbit

* **New Features**
  * Added “active session near bottom” tracking driven by scroll position, enabling the viewer to react as users approach the end of the feed.
* **Improvements**
  * File-watching now primes signatures on startup and suppresses duplicate change emissions, so updates occur only on real content changes.
  * Session refresh and project refresh handling is now coalesced/debounced to reduce redundant reloads.
  * In-place session reloads preserve existing message state and skip updates when refreshed messages are equivalent.
* **Tests**
  * Expanded coverage for signature priming, refresh scheduling/coalescing, and message reload regressions.


bugsyyu:
@coderabbitai review

PR #368, opened 2026-06-13
fix(rename): use Claude native rename event format

bugsyyu:
## Summary
- Make Claude native rename append the same `system/local_command` event shape produced by Claude Code `/rename`
- Keep the first user message intact instead of prefix-mutating conversation history
- Reset Claude native names by removing rename events and cleaning up legacy `[Title]` prefixes
- Update the rename dialog/session state so Claude renamed sessions use standalone titles and `is_renamed` stays in sync

## Root Cause
The previous implementation renamed Claude sessions by editing the first user message to `[Title] original message`. However, Claude Code `/rename` writes a separate `system` message with `subtype: "local_command"` and `<local-command-stdout>Session renamed to: ...</local-command-stdout>`. The loader already treats only that event format as an explicit native rename, so viewer-created renames were not equivalent to Claude Code renames and still loaded with `is_renamed: false`.

## Validation
- `cargo test --manifest-path src-tauri/Cargo.toml --features webui-server commands::session::rename::tests`
- `cargo test --manifest-path src-tauri/Cargo.toml --features webui-server commands::session::load::tests::test_should`
- `cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets --all-features -- -D warnings`
- `corepack pnpm build`
- `corepack pnpm test --run src/test/SessionItem.test.tsx src/test/useSessionEditing.test.tsx`
- WebUI API smoke test against a temporary Claude JSONL file: rename wrote a `system/local_command` event and loaded with `is_renamed: true`; reset removed the event and loaded with `is_renamed: false`

## Summary by CodeRabbit

* **New Features**
  - Rename dialog now returns whether a native rename occurred and shows provider-specific previews (Claude included).
  - Session list syncs rename status so UI reflects native renames immediately.

* **Bug Fixes**
  - Renames are normalized/validated (trim, no newlines) and recorded robustly.
  - Reset now removes native rename metadata and legacy title prefixes consistently.


jhlee0409:
Thanks @bugsyyu — the rename-format fix itself looks good. Two things before it can merge: (1) this PR also bundles `README.md` + the 4 translated READMEs + `CHANGELOG.md`. Per the repo's branch strategy, docs/README changes land only in release commits, not in feature PRs — these look like a stale v1.13.0 docs commit picked up from `main`. Please drop them so the diff is just the rename change. (2) Please rebase onto the latest `develop` (it now conflicts after #353/#367 merged). Thanks!

PR #372, opened 2026-06-13
fix(session): use in-app delete confirmation

bugsyyu:
## Summary
- Replace the WebUI browser confirmation prompt with an in-app session delete dialog that matches the existing site UI.
- Keep deletion behind an explicit double-check: menu actions only open the dialog, and the API call runs only after the destructive confirmation button is clicked.
- Remove the unconditional Tauri dialog call from the delete flow so WebUI no longer crashes on missing `invoke`.
- Add regression coverage for opening/cancelling/confirming deletion and for the new dialog component.

## Root cause
The session delete flow imported `@tauri-apps/plugin-dialog` and called `ask()` unconditionally. In WebUI/browser mode the dialog plugin tries to call Tauri `invoke`, which is unavailable, so deletion failed before the confirmation could complete.

## User impact
WebUI users now get a styled confirmation dialog instead of a browser prompt, and session deletion works without the `Cannot read properties of undefined (reading invoke)` error.

## Validation
- `corepack pnpm exec eslint src/components/SessionItem/SessionItem.tsx src/components/SessionItem/components/SessionDeleteDialog.tsx src/components/SessionItem/hooks/useSessionEditing.ts src/test/useSessionEditing.test.tsx src/test/SessionDeleteDialog.test.tsx`
- `corepack pnpm exec vitest run src/test/useSessionEditing.test.tsx src/test/SessionDeleteDialog.test.tsx`
- `corepack pnpm build`

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated session deletion confirmation dialog with provider-specific title/description and clear destructive delete action.
* **Bug Fixes**
  * Improved the deletion flow to use a two-step confirmation; the dialog can’t be dismissed while deletion is in progress, and canceling prevents deletion.
* **Tests**
  * Expanded coverage to assert the deletion API call, dialog open/close behavior, selection reset after deletion, and success feedback.
  * Added unit tests for the delete dialog’s rendering and cancel/confirm interactions.

### DIFF
--- a/src/components/SessionItem/SessionItem.tsx
+++ b/src/components/SessionItem/SessionItem.tsx
@@ -5,6 +5,7 @@ import { useSessionEditing } from "./hooks/useSessionEditing";
 import { SessionHeader } from "./components/SessionHeader";
 import { SessionNameEditor } from "./components/SessionNameEditor";
 import { SessionContextMenu } from "./components/SessionContextMenu";
+import { SessionDeleteDialog } from "./components/SessionDeleteDialog";
 import { SessionMeta } from "./components/SessionMeta";
 import type { SessionItemProps } from "./types";
 import type { Boundary } from "@/utils/contextMenu";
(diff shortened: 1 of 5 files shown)


jhlee0409:
Closing as a duplicate of #373. #373 is a strict superset of this PR — it ships the same in-app delete confirmation dialog (the `SessionDeleteDialog` component and its test) **plus** Codex CLI rename/delete support, and both edit the same `useSessionEditing` code path. Consolidating review in #373 to avoid conflicting changes to the same files. Thanks @bugsyyu — the work lives on in #373, which has been retargeted from `main` to `develop` per the repo's branch strategy.

PR #374, opened 2026-06-14
fix(server): guard unsafe no-auth binds

bugsyyu:
## Summary

Adds a startup safety guard for WebUI server mode when authentication is disabled.

- Refuses `--no-auth` on network-reachable bind hosts unless `--allow-unsafe-no-auth` is explicitly provided.
- Keeps `--no-auth` usable for local-only loopback hosts such as `127.0.0.1`, `::1`, and `localhost`.
- Warns when a custom CLI/env token is shorter than 32 characters.
- Updates README and server guide CLI/auth notes across existing localized docs.

## Why

The server currently defaults to `0.0.0.0`, so `--serve --no-auth` can expose conversation history to the local network. This change keeps local development convenient while making unauthenticated network exposure an explicit opt-in.

## Validation

- `cargo test --features webui-server webui_server_startup_tests --lib`
- `cargo test --features webui-server --lib -- --test-threads=1`
- `cargo clippy --features webui-server --all-targets -- -D warnings`
- `cargo run --features webui-server -- --serve --no-auth` exits with code 2 and the new refusal message

Note: a default parallel `cargo test --features webui-server --lib` run exposed existing archive/settings test isolation failures; the same full lib suite passes with `--test-threads=1`.


jhlee0409:
Folded into #384. #382's branch already ships the identical non-loopback `--no-auth` guard (`validate_auth_startup_options` + `--allow-unsafe-no-auth`), so this behavior is covered there (and called out as a BREAKING change in the #384 description for release notes). Credited via Co-Authored-By. Thanks @bugsyyu!

PR #375, opened 2026-06-14
feat(server): add read-only mode

bugsyyu:
## Summary

Adds a WebUI server read-only mode for shared viewers.

- Adds `--read-only` to server mode.
- Exposes `/api/get_server_config` so the frontend can detect read-only mode.
- Returns `403` for mutating WebUI API endpoints, including session rename/delete, file restore/export writes, metadata/settings saves, presets, MCP saves, archive mutations, screenshots, and feedback submission.
- Disables common frontend write controls for session actions, settings/custom directories, WSL settings, and archive mutations.
- Updates README and server guide CLI option tables.

## Why

When the WebUI is shared with family members, teammates, or over a reverse proxy, users may want browsing and search without allowing accidental local changes such as deleting sessions, renaming conversations, changing settings, or modifying archives.

## Validation

- `npx vitest run src/test/archiveSlice.test.ts src/test/serverSlice.test.ts src/test/useSessionEditing.test.tsx src/test/SessionContextMenu.test.tsx src/test/metadataSlice.test.ts`
- `npx tsc --noEmit`
- `npm run build`
- `cargo test --features webui-server --lib -- --test-threads=1`
- `cargo clippy --features webui-server --all-targets -- -D warnings`
- `cargo build --features webui-server`
- Manual HTTP check: `--read-only` returns `{ "readOnly": true }` from `/api/get_server_config` and `403` from `/api/delete_session`.


jhlee0409:
Consolidated into #384. Read-only mode is adopted there, re-implemented as **deny-by-default** — an allowlist of read-safe routes rather than a blocklist of mutations, so a newly added mutating route is blocked automatically instead of leaking through. (A self-maintaining test caught 7 read routes a naive blocklist→allowlist flip would have wrongly blocked.) Credited via Co-Authored-By. Thanks @bugsyyu!

PR #377, opened 2026-06-14
feat(server): add WebUI cookie auth

bugsyyu:
## Summary
- add /api/auth/login and /api/auth/logout for HttpOnly WebUI auth cookies
- allow protected API/SSE requests to authenticate via Bearer header, auth cookie, or existing SSE query token fallback
- exchange browser URL tokens for HttpOnly cookies during startup, then remove the readable localStorage token
- keep existing Bearer/localStorage behavior as fallback when cookie login fails
- update authentication docs across README/server guide variants

## Verification
- npx vitest run src/utils/platform.test.ts
- npx tsc --noEmit
- npm run build
- cargo test --features webui-server server::tests --lib
- cargo test --features webui-server --lib -- --test-threads=1
- cargo clippy --features webui-server --all-targets -- -D warnings
- cargo build --features webui-server
- pre-commit: eslint, cargo fmt --check, cargo clippy --all-features
- manual smoke: verified unauthenticated API 401, /api/auth/login Set-Cookie, cookie-authenticated API 200, cookie-authenticated SSE 200, /api/auth/logout cookie clear, wrong-token login 401


jhlee0409:
Superseded by #384 (built on #382's account auth). Account auth delivers the HttpOnly-cookie + CSRF hardening as a sub-case of full session auth, so a standalone cookie-auth layer is redundant rather than buggy — your approach informed the token-compat cookie path. Credited via Co-Authored-By on #384. Thanks @bugsyyu!

PR #381, opened 2026-06-14
perf(session): paginate project session loading

bugsyyu:
## Summary
- Add a cache-aware paginated session loading API for project session lists.
- Load the first page of sessions immediately and append more sessions as the sidebar list is scrolled.
- Keep the existing full session loading APIs intact for compatibility.

## Why
Projects with thousands of session files currently wait for every session to be processed and serialized before the sidebar can show anything. This makes expanding large projects feel stalled even though only the first visible rows are needed initially.

## Implementation notes
- Claude projects use cached metadata and file mtimes to order candidates without parsing every JSONL file up front.
- Changed or uncached files are parsed only when they fall into the requested page.
- Provider session loading exposes the same paginated response shape; non-Claude providers retain their existing loaders and paginate the result.
- The sidebar keeps virtualized rendering and requests the next page near the bottom of the list.

## Validation
- `npm run build`
- `npm test -- src/components/ProjectTree/components/__tests__/SessionList.test.tsx --run`
- `cargo test --manifest-path src-tauri/Cargo.toml --features webui-server test_load_project_sessions_page_uses_cache_and_offsets -- --nocapture`
- pre-commit `eslint`, `cargo fmt --check`, and `cargo clippy --all-targets --all-features -- -D warnings`

## Privacy check
The PR branch was created from upstream `main` and contains only this performance commit. The diff was checked for deployment IPs, tokens, private hostnames, and local machine paths before opening the PR.

## Summary by CodeRabbit

* **New Features**
  * Added paginated session loading for project and cross-provider session lists, including recency-based ordering and “next page” metadata.
  * Added a “Load more” sessions UI with count/progress, loading feedback, and automatic fetching as you scroll.
  * Introduced a shared paginated `SessionPage` response type and registered the new endpoints/commands (including read-only access).
* **Tests**
  * Extended pagination, ordering, cache-reuse behavior, invalid path/candidate handling, missing/invalid data resilience, and symlink-outside-root filtering coverage.


jhlee0409:
Thanks @bugsyyu! CI is currently red on this branch — the new test `commands::session::load::tests::test_load_project_sessions_page_skips_invalid_candidates` **FAILS** in the nextest run, so this can't merge yet. Could you take a look at that case? Also, `develop` has advanced since you opened this (#353 CodeBuddy + #367 watcher fix landed), so please rebase onto the latest `develop` — it currently conflicts. (Note: this PR was retargeted from `main` to `develop` per the repo's branch strategy.)

jhlee0409 (changes requested):
Thanks for this, @bugsyyu — the shape is right: a cache-aware paginated backend command with Tauri+Axum parity handled from day one, and the Rust-side tests for ordering, symlink skipping, and path validation are well-structured. A few things are blocking merge, though:

**1. CI test failure (blocker).** `test_load_project_sessions_page_skips_invalid_candidates` fails in the CI run (panic at `load.rs:2482`, a `total`-count assertion). The cause looks like `newly_dropped_candidates` only accumulating within a single call, so `total` on page 2 is computed without knowledge of what page 1 dropped — making `total` non-monotone across pages when there are corrupt/unreadable JSONL files. The test is a good one; the implementation needs to match it.

**2. Race condition in `loadMoreSessions` (blocker).** In `projectSlice.ts`, `loadMoreSessions` uses `getRequestId("selectProject")` instead of `nextRequestId(...)`, so it piggybacks on the latest counter without incrementing it. If the user switches projects while a `loadMore` is in flight, the guard `if (requestId !== getRequestId(...))` never fires and the wrong project's sessions get appended to the new project's list. Either use `nextRequestId`, or snapshot `selectedProject.path` before the await and compare after.

**3. Missing i18n keys (blocker for `i18n:validate`).** `t("session.loadMore", "Load more")` is used in `SessionList.tsx` but the key is absent from all five locale files. Please add `session.loadMore` to `src/i18n/locales/{en,ko,ja,zh-CN,zh-TW}/session.json` and run `pnpm run i18n:validate`.

**4. Rebase needed.** `develop` has moved a fair bit since this was opened (it's now at v1.16.0), so this conflicts on `CHANGELOG.md`/`README.md`. Please rebase onto the current `develop` tip and drop the CHANGELOG/README churn from the old base (it has already landed).

**5. Design gap worth a note.** `messageSlice.refreshSession` still calls the full `load_project_sessions` and then `setSessions(...)`, which resets `hasMoreSessions: false`. So after any file-change-triggered refresh, the paginated state is overwritten with a full load — defeating the perf goal on refresh for large projects. A follow-up is fine, but please add a `TODO` in `refreshSession` so it isn't forgotten.

Once those are addressed I'm happy to take another look — the backend pagination + parity are solid and this is worth getting right. Thanks again!

jhlee0409:
Following up on blocker #1 (the CI failure) with the precise root cause — traced through the CI log and the diff.

**Failure:** `test_load_project_sessions_page_skips_invalid_candidates` panics at `load.rs:2482` — `assertion left == right failed; left: 3, right: 2`. So `second_page.total` is **3 when the test expects 2**.

**Setup:** 3 candidate files — `valid-1.jsonl`, `invalid.jsonl` (`{}` — parses as JSON but isn't a valid session), `valid-2.jsonl`.

**`total` is computed per-call:**

```rust
let total = candidate_total
    .saturating_sub(known_dropped_candidates + newly_dropped_candidates);
```

`newly_dropped_candidates` only counts invalid candidates encountered by *this call's* scan loop.

**Why page 2 breaks:**

- **Page 1** (`offset=0, limit=2`): the loop scans all 3 candidates — valid-1 (valid), invalid (`newly_dropped=1`), valid-2 (valid, page now full). `total = 3 − (0 + 1) = 2` ✅
- **Page 2** (`offset=next_offset=3, limit=2`): `while sessions.len() < limit && next_offset(3) < candidate_total(3)` is immediately false, so the scan loop never runs → `newly_dropped = 0`, and the pre-scan `known_dropped` doesn't account for the invalid candidate that page 1 dropped. `total = 3 − (0 + 0) = 3` ❌ → `left: 3, right: 2`.

**Root cause:** `total` is derived only from the candidates *the current call examined*, so it's non-monotone across pages — 2 on page 1, 3 on page 2. To return an exact `total` you'd need to know how many candidates are invalid independently of which page is served — e.g. scan/parse all candidates once up front to count valid sessions, persist the dropped-candidate set across calls, or drop the exact `total` in favor of a `has_more`-only model.

(The passing `..._total == 3` test has no invalid candidates, so `candidate_total == valid count` and the bug doesn't surface — `skips_invalid_candidates` is the one that correctly catches it. The test is right; the implementation needs to match it.)

jhlee0409:
Landed! Since the branch had gone quiet, I took it over to get it shipped — the pagination design here was solid and survives intact. What was done on top:

- **Merged a month of develop** (multi-select sessions, three new providers, codex loader changes) — your request-id guard in `selectProject` now coexists with the newer selection-mode reset.
- **Fixed the failing `total` accounting**, which turned out to have two mechanisms: (1) the paged loader keyed its cache by canonicalized paths while the full loader uses raw paths, so on any symlinked path (`/var` → `/private/var` on macOS — including every TempDir test) cache lookups always missed and earlier pages' dropped candidates were forgotten; (2) the disk cache write is best-effort, so a failed save also lost the accounting. Cache keys now use raw paths (canonical root kept only for the symlink-escape check), and a small (mtime,size)-validated in-memory registry keeps dropped-candidate accounting cumulative even without the disk cache — with a regression test that deletes the cache file between pages.
- **Read-only server guard**: classified both new pagination routes in `READ_ONLY_ALLOWED_API_PATHS`.
- **i18n**: the "Load more" button now has keys in all 5 locales.

CI fully green (816+ backend / 903 frontend). Known follow-ups noted for later: refresh-all snaps a deep-scrolled list back to page 1, and non-Claude providers get correct-but-not-faster pagination. Thanks @bugsyyu for the contribution — the core design is yours. Merging.

PR #417, opened 2026-06-21
feat(providers): add Amazon Q Developer CLI provider

jhlee0409:
## Summary

Adds the **Amazon Q Developer CLI** (`q chat`) as a provider, reading its local SQLite chat store. Brings supported provider IDs to **20**.

- **`amazon_q.rs`** — reads `dirs::data_local_dir()/amazon-q/data.sqlite3`, table `conversations(key, value)` where `key` = cwd and `value` = serialized `ConversationState`. v1 schema stores **one conversation per cwd**, so each `key` is a project with a single session; session times are derived from the history's user timestamps (v1 has no `created_at`/`updated_at`).
- **`q_conversation.rs`** (new shared module) — the Amazon Q CLI `conversations` table and Kiro CLI's `conversations_v2` store the **identical** `value` shape (verified from `aws/amazon-q-developer-cli` source: same `ConversationState`, same externally-tagged `Prompt`/`ToolUseResults`/`CancelledToolUses`/`Response`/`ToolUse`). The value→message conversion now lives here, parameterized by provider id.
- **`kiro.rs`** refactored to use the shared converter (drops ~120 duplicated lines) and now resolves its DB via `dirs::data_local_dir()` — **fixing a latent bug**: the old code hardcoded the Windows path to `AppData\Roaming\kiro-cli`, but upstream uses `data_local_dir()` (`%LOCALAPPDATA%`, i.e. **Local**).

## Verification

- Table + `value` schema and the v1↔v2 difference confirmed from upstream Rust source (migration `007_conversations_table.sql`, `ConversationState`/`message.rs`). v1↔v2 differ only at the row/column layer; the `value` JSON is the same type → converter reuse is safe.
- New providers tested with in-memory SQLite DBs; the shared converter has unit tests (incl. `CancelledToolUses`, tool-name mapping, time bounds).
- Gates: `tsc` ✅ · `vitest` 864 ✅ · `eslint` ✅ · `i18n:validate` ✅ · `cargo test` 727 ✅ · `clippy -D warnings` ✅ · `fmt` ✅

## Notes

- ⚠️ Not yet runtime-verified against a real Amazon Q install — a real-file sanity check is recommended before release.
- The post-rebrand **Kiro CLI** (`kiro-cli/data.sqlite3`, `conversations_v2`) is already covered by the existing `kiro` provider; this adds the legacy/standalone **`amazon-q`** store.

## Summary by CodeRabbit

* **New Features**
  * Added Amazon Q CLI as a selectable provider, including project/session browsing and cross-provider search.
  * Updated provider name/badge display with new translations (English, Japanese, Korean, Simplified Chinese, Traditional Chinese).
  * Amazon Q provider sessions now participate in real-time session updates and show up in provider project counts.

### DIFF
--- a/src-tauri/src/commands/multi_provider.rs
+++ b/src-tauri/src/commands/multi_provider.rs
@@ -46,6 +46,7 @@ pub async fn scan_all_projects(
             "cursor".to_string(),
             "cursor-agent".to_string(),
             "aider".to_string(),
+            "amazonq".to_string(),
             "antigravity".to_string(),
             "codebuddy".to_string(),
             "kiro".to_string(),
@@ -226,6 +227,16 @@ pub async fn scan_all_projects(
             Ok(projects) => all_projects.extend(projects),
             Err(e) => {
                 log::warn!("Aider scan failed: {e}");
+            }
+        }
+    }
+
+    // Amazon Q Developer CLI — amazon-q/data.sqlite3 (conversations table)
+    if providers_to_scan.iter().any(|p| p == "amazonq") {
+        match providers::amazon_q::scan_projects() {
+            Ok(projects) => all_projects.extend(projects),
+            Err(e) => {
+                log::warn!("Amazon Q scan failed: {e}");
             }
         }
     }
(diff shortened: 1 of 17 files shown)


jhlee0409:
Thanks @coderabbitai — valid finding, fixed in 55e47e8. `load_sessions` now uses `.optional().map_err(...)?` so a missing project key still returns an empty list, but genuine DB read/schema errors propagate (matching `load_messages`). `cargo test` 727 · `clippy -D warnings` · `fmt` green locally.

PR #431, opened 2026-06-30
feat(sidebar): collapsible provider filter panel

asukaneko:
## Summary

- Make the project sidebar provider filter collapsible so the session list keeps more vertical space by default.
- Keep a compact provider summary visible while collapsed, including the current filtered project count and reset action.
- Increase the expanded provider filter scroll area slightly and prevent session timestamps from wrapping vertically in narrow sidebars.

## Why

The provider list could consume too much of the left sidebar, leaving the session list cramped. In narrow layouts, session relative times could wrap character-by-character.

## Validation

- `corepack pnpm run lint`
- `corepack pnpm run build`

## Summary by CodeRabbit

* **New Features**
  * Provider filters in the project tree can now be collapsed and expanded for a more compact view.
  * The provider filter header displays a summary of selected providers and an aggregate count.

* **Bug Fixes**
  * The provider filter panel open/closed state is persisted between visits.
  * Improved spacing and layout behavior for session metadata badges and details to reduce wrapping/clipping issues.

### DIFF
--- a/src/components/ProjectTree/index.tsx
+++ b/src/components/ProjectTree/index.tsx
@@ -860,11 +860,9 @@ export const ProjectTree: React.FC<ProjectTreeProps> = ({
                     "hover:bg-accent/8 hover:text-accent focus:outline-none focus:ring-1 focus:ring-accent/40"
                   )}
                   title={providerFilterSummary}
-                  aria-label={
-                    areProviderFiltersOpen
-                      ? t("common.collapse", "Collapse")
-                      : t("common.expand", "Expand")
-                  }
+                  aria-label={`${areProviderFiltersOpen
+                    ? t("common.collapse", "Collapse")
+                    : t("common.expand", "Expand")}: ${providerFilterSummary} (${providerFilterCount})`}
                 >
                   {areProviderFiltersOpen ? (
                     <ChevronDown className="h-3 w-3 shrink-0" />
--- a/src/components/ProjectTree/index.tsx
+++ b/src/components/ProjectTree/index.tsx
@@ -3,6 +3,8 @@ import React, { useCallback, useEffect, useMemo, useRef, useState } from "react"
 import {
   Folder,
   Database,
+  ChevronDown,
+  ChevronRight,
   List,
   FolderTree,
   GitBranch,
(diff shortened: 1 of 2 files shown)


jhlee0409 (changes requested):
Thanks for this, @asukaneko — the collapsible filter panel is a genuine usability improvement for narrow sidebars, and the implementation is clean: localStorage access is guarded with try/catch, the lazy state initializer is used correctly, all i18n keys (`common.collapse`, `common.expand`, `session.board.controls.all`) are pre-existing across all five locales, and the trigger has a proper dynamic `aria-label`.

A couple of things to address before we can merge:

**Required**
1. **PR title**: the `[codex]` prefix is the AI tool name, not a conventional-commit type. Please rename it to something like `feat(sidebar): collapsible provider filter panel` so it fits the project's commit-message convention.

**Please confirm**
2. **Default-collapsed on first visit**: `loadProviderFiltersOpenState` returns `false` when there is no saved preference, so the provider-filter panel starts hidden for every new user. That's fine as an intentional choice, but it could be surprising for someone who has a non-default provider filter active and doesn't notice the chevron. Can you confirm collapsed-by-default is the intended first-launch experience? If you'd prefer "open once, then respect the saved preference", a one-line change (e.g. `localStorage.getItem(...) !== "false"`) would do it.

**Minor nit (not a blocker)**
- The reset button is now icon-only. It keeps a correct `aria-label` so accessibility is fine, and the `title` tooltip is still there — just flagging in case you'd prefer to keep the visible text label for discoverability.

Once the title is updated and you've confirmed the collapsed-default intent, this is good to go. Thanks!

asukaneko:
Thanks for the thorough review, @jhlee0409!
1. PR title — Updated to feat(sidebar): collapsible provider filter panel.
2. Default-collapsed intent — Confirmed, this is intentional. The collapsed trigger still surfaces the active filter context: it renders providerFilterSummary (truncated) and the providerFilterCount badge, so a user with a non-default provider selection sees their active filters at a glance without expanding. Keeping it collapsed by default aligns with the PR's goal of reclaiming vertical space for the session list on narrow sidebars. If the first-launch surprise turns out to be a real issue in practice, the !== "false" flip is a trivial follow-up.
3. Reset button — Keeping it icon-only. It already has both aria-label and title for discoverability, and an icon button fits the compact sidebar better than a labeled one. Happy to revisit if it reads as unclear to users.

jhlee0409 (approved):
Both required items resolved: title renamed to conventional-commit form, and collapsed-by-default confirmed intentional (the collapsed trigger still surfaces the active filter summary + count badge). Verified the merge-into-develop result locally — tsc, eslint, i18n:validate, and vitest (868) all green. Thanks @asukaneko!

PR #432, opened 2026-06-30
fix(providers): split opencode global sessions by directory

asukaneko:
## What & why

  OpenCode can store some sessions under the special `global` project with `project.worktree = "/"` and no project name.
  The viewer previously derived the project label from `worktree`, so those sessions collapsed into an `unknown` project
  with no useful path.

  This changes OpenCode SQLite project scanning to split the `global` project by each session's `directory`, using
  stable virtual project IDs while preserving the real session paths (`opencode://global/<session_id>`) for message
  loading and session operations.

  Closes #

  ## Type

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor / perf
  - [ ] Docs

  ## Checklist

  - [ ] `pnpm exec tsc --build .` and `pnpm lint` pass
  - [ ] **i18n**: any new user-facing string is `t()`-wrapped and the key exists in **all 5 languages** (`en, ko, ja,
  zh-CN, zh-TW`) — verified with `pnpm run i18n:validate`

  - [ ] Registered in **both** the Tauri `generate_handler!` (`src-tauri/src/lib.rs`) **and** the Axum WebUI router
  (`src-tauri/src/server/mod.rs`) — otherwise `--serve` mode 404s

  - [ ] Followed the existing provider pattern in `src-tauri/src/providers/`
  - [ ] Session discovery verified on macOS / Linux / Windows (and WSL if applicable)
  - [ ] No leftover identifiers copied from another provider (names, comments, fixtures)

## Summary by CodeRabbit

* **New Features**
  * Added support for virtual global-directory projects, grouping and displaying sessions by their originating directory.
* **Bug Fixes**
  * Improved project matching so sessions are associated with the correct project scope.
  * Enhanced session loading and merging to better prevent duplicates and ensure directory-scoped behavior works correctly, including for global-derived sessions.

### DIFF
--- a/src-tauri/src/providers/opencode.rs
+++ b/src-tauri/src/providers/opencode.rs
@@ -8,6 +8,9 @@ use std::collections::HashSet;
 use std::fs;
 use std::path::{Path, PathBuf};
 
+const OPENCODE_GLOBAL_PROJECT_ID: &str = "global";
+const OPENCODE_GLOBAL_DIRECTORY_PREFIX: &str = "global-dir-";
+
 /// Convert epoch milliseconds to RFC 3339 string
 fn epoch_ms_to_rfc3339(ms: u64) -> String {
     #[allow(clippy::cast_possible_wrap)]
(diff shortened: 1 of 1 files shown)


jhlee0409 (changes requested):
Thank you for this fix — the `global`-to-per-directory splitting design is clean and the `OpenCodeProjectRef` abstraction is a good addition.

**One must-fix bug before we can merge:** sessions with an empty `directory` column (`''`, which is the OpenCode schema default) are mishandled.

- During **scanning**, `scan_global_directory_projects` uses `COALESCE(NULLIF(s.directory, ''), ?)` so an empty-directory session is bucketed under the worktree path `"/"`, and the virtual project ID becomes `stable_directory_hash("/")` (`af63a24c860189fe`).
- During **session loading**, `load_sessions_from_db` reads the raw column value `""` and calls `matches_directory("")` → `stable_directory_hash("")` (`cbf29ce484222325`) — a different hash.

The two hashes don't match, so those sessions pass the scan (they show in the project's session count) but are silently dropped at load time — the project opens with zero messages. The OpenCode schema declares `directory TEXT NOT NULL DEFAULT ''`, so this affects any global session inserted without a directory value.

**Suggested fix:** remove the `COALESCE` from the SQL `GROUP BY` and use the raw `s.directory` value as the hash key, so scanning and loading hash the same value. The worktree fallback for the *display name* is already handled by the `if directory.trim().is_empty() { project.worktree.clone() }` branch in the `.map()` closure, so that can stay as-is. Please also add a test that inserts a global session with `directory = ''` and asserts it appears in the virtual project's session list after loading.

Everything else looks correct: path parsing is properly centralized, the FNV-1a hash is deterministic and dependency-free, Windows backslash normalization is present, the message-loading path is unaffected (the virtual `global-dir-*` IDs are never written into `file_path`), and the existing tests are correctly updated. This is a net improvement over the status quo once the empty-directory case is addressed. Thanks again!

asukaneko:
Thanks for catching this, @jhlee0409 — good catch on the hash mismatch.

**Fix pushed.** Removed the `COALESCE(NULLIF(s.directory, ''), ?2)` from both the `SELECT` and `GROUP BY` in `scan_global_directory_projects`, so scanning now groups by the raw `s.directory` value — the same input `load_sessions_from_db` feeds into `matches_directory`. The worktree fallback for the display `actual_path` is untouched (still handled by the `if directory.trim().is_empty() { project.worktree.clone() }` branch), so empty-directory sessions now bucket under `stable_directory_hash('')` consistently in both phases and load correctly.

**Test added:** `sqlite_load_sessions_includes_global_empty_directory_session` inserts a global session with `directory = ''` alongside a normal `/tmp/a` one, scans the projects, finds the empty-directory virtual project by its `actual_path == /` fallback, and asserts that `load_sessions_from_db` returns the session (not silently dropped). Verified locally with `cargo test --lib providers::opencode -- --test-threads=1` — all 14 opencode tests pass, including the new one.

Everything else in the review is already addressed (path parsing centralized in `OpenCodeProjectRef`, FNV-1a hash is deterministic and dependency-free, Windows backslash normalization present, virtual `global-dir-*` IDs never flow into `file_path`). Ready for another look.

jhlee0409 (approved):
Re-reviewed: the empty-`directory` data-loss bug is fixed — `scan_global_directory_projects` now groups by raw `s.directory` (COALESCE removed from SELECT + GROUP BY), so scan and load hash the same value and empty-directory global sessions load correctly. Regression test `sqlite_load_sessions_includes_global_empty_directory_session` added. Rust Test Suite green on the PR. Thanks @asukaneko!

PR #439, opened 2026-07-01
feat(web): add global conversation refresh

asukaneko:
﻿## What & why

Adds a global conversation refresh action in the header so users can manually rescan all conversation projects, reload the current project/session when still present, and clear stale selections when projects or sessions disappear.

This complements the existing current-session refresh and file watcher behavior with an explicit whole-app refresh path.

Closes #

## Type

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / perf
- [ ] Docs

## Checklist

- [x] **i18n**: any new user-facing string is `t()`-wrapped and the key exists in **all 5 languages** (`en, ko, ja, zh-CN, zh-TW`) — verified with `pnpm run i18n:validate`

- [x] N/A — no frontend-callable backend command added

- [x] N/A — no new provider added

### DIFF
--- a/src/i18n/locales/en/session.json
+++ b/src/i18n/locales/en/session.json
@@ -191,6 +191,7 @@
   "session.nativeRename.warningForgeCode": "This operation updates the ForgeCode conversation title stored in the Forge database.",
   "session.nativeRename.warningOpenCode": "This operation modifies the OpenCode session metadata file. The change is reversible.",
   "session.refresh": "Refresh Sessions",
+  "session.refreshAllConversations": "Refresh all conversations",
   "session.rename": "Rename session",
   "session.renameAction": "Rename session",
   "session.renameHint": "Double-click to rename",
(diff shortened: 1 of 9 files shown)


jhlee0409:
Thanks for this — nice, well-scoped addition. I checked the diff against the repo's usual failure modes (Tauri/Axum command parity, i18n, error swallowing, stale-selection handling) and everything checks out:

- No new backend command was added; `refreshAllConversations` composes existing store actions, so there's no Tauri/Axum parity risk.
- Errors are surfaced via `toast.error` + `setError`, not swallowed.
- The new i18n key is present in all 5 locales and `i18n:validate` passes locally.
- Stale-project/session clearing reuses the app's existing `clearProjectSelection` pattern and is covered by 4 new unit tests, which pass.

One note for the record (non-blocking): `refreshAllConversations` intentionally skips `detectProviders()`, so a provider CLI installed after app launch needs an app restart to be picked up — reasonable scope for a lightweight rescan; can be a follow-up if anyone asks.

LGTM — merging.

PR #450, opened 2026-07-08
fix(project): count only top-level jsonl as sessions in scan_projects

ruskin-pkucala:
## Problem

The session count shown in the project tree is inflated and diverges from the count shown when a project is opened.

Example: a project displays **530** sessions in the tree, but opening it lists only **73**.

## Root cause

`scan_projects` walks the project directory recursively (`WalkDir` without depth limit) and increments `session_count` for **every** `*.jsonl`, including sidechain/subagent files that Claude Code stores under `<project>/<session-uuid>/subagents/*.jsonl`.

`load_project_sessions`, on the other hand, filters those out when `exclude_sidechain=true` (the frontend default), so it returns only the top-level main sessions. The two functions disagree on "what counts as a session", and the tree number is the inflated one.

## Fix

Only count top-level `*.jsonl` (relative path with a single component) as sessions:

```rust
let is_direct_session = jsonl_entry.path()
    .strip_prefix(entry.path())
    .is_ok_and(|relative| relative.components().count() == 1);
if is_direct_session {
    session_count += 1;
}
```

The recursive `WalkDir` is preserved on purpose, so `last_modified`, `message_count`, and the cwd fallback candidates keep their existing behavior — only `session_count` is narrowed.

This relies on Claude Code's storage convention that subagent jsonl always live in subdirectories (`<uuid>/subagents/`), never at the top level (see `utils::find_subagent_files`). A comment is added at the call site to record this assumption.

## Known follow-up (not addressed here)

`message_count` still accumulates file sizes from sidechain files, so it remains inconsistent with the opened-session view. Intentionally left for a separate change to keep this PR focused; flagged here so reviewers aren't surprised.

## Tests

Adds `tests/session_count_test.rs` with three cases using a temp-dir fixture (no local Claude data required, CI-friendly):

- `sidechain_files_are_not_counted_as_sessions` — 2 top-level + 3 sidechain in `<uuid>/subagents/` → count is 2
- `only_top_level_sessions_are_counted` — 4 top-level + 10 sidechain → count is 4
- `project_with_no_top_level_sessions_is_skipped` — a dir with only sidechain files has no main session and is skipped (`session_count == 0`)

```
running 3 tests
test session_count_tests::project_with_no_top_level_sessions_is_skipped ... ok
test session_count_tests::sidechain_files_are_not_counted_as_sessions ... ok
test session_count_tests::only_top_level_sessions_are_counted ... ok
test result: ok. 3 passed; 0 failed
```

## Verification

On real data, a project that previously showed 530 in the tree now shows 73, matching the top-level `*.jsonl` count and the opened-session list.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated project scanning so session counts only include top-level `*.jsonl` files, excluding nested session data.
  * Projects that contain only nested session files are no longer returned by the scan.
* **Tests**
  * Added regression coverage for top-level vs nested `*.jsonl` session counting and for projects omitted when they contain only nested data.
* **Documentation**
  * Added the `ZCode` provider label to English and all supported locales (JP/KR/ZH-CN/ZH-TW).


ruskin-pkucala:
Thanks for catching that! You're right, the existing tests' assertions needed to be updated to reflect the new top-level-only counting logic.

I've updated the expected session_count from 2 to 1 in both `test_scan_projects_prefers_top_level_cwd_over_subagent_cwd` and `test_scan_projects_nested_sessions` (commit 75fac06).

Regarding the OS path separator concern, Rust's `std::path::components()` handles path normalization cross-platform, so it should work correctly on both Windows and Unix. All tests are passing locally now.

jhlee0409:
Heads-up: your Rust CI failure was **not caused by this PR**. CI's stable Rust moved to 1.97, whose clippy flags two pre-existing patterns in the base branch (`stats.rs`, `kimi.rs`). That's now fixed on `develop` (089c0be).

Please rebase (or merge `develop` into your branch) and push — your own Test Suite already passed, so this should go fully green. Note the PR base was also switched to `develop` per the repo's branch strategy.

jhlee0409:
All green after the branch update (Frontend + Rust suites on 2fce1eb, including your 3 session-count regression tests). The fix is exactly right — counting only top-level `*.jsonl` in `scan_projects` matches what `load_project_sessions` shows, eliminating the 530-vs-73 divergence. I pushed two small maintenance commits to your branch along the way (develop merge + a clippy 1.97 doc-lint backtick fix in the new test file). Thanks for the contribution — merging.

PR #456, opened 2026-07-10
feat(sessions): multi-select mode with mass delete

tonysurfly:
## Summary

Adds a **multi-select mode** to the project session list so several sessions can be selected and acted on at once — the two main operations being **mass delete** and **copy session IDs**.

Targets `develop` per the branch strategy in `CLAUDE.md`.

## Interaction

- A **"Select" toggle** (list-checks icon) in the session-list controls enters selection mode; **Esc** or the bar's ✕ exits.
- In selection mode each row shows a **checkbox**: plain click toggles it, **Shift+click** selects a range from the anchor, **Cmd/Ctrl+click** toggles individually.
- A sticky **action bar** shows: `N selected` · **Select all / Clear** · **Copy IDs** · **🗑 Delete (N)**.
- Outside selection mode, clicking a session opens it exactly as before — no change to existing behavior.

## Implementation notes

- **No new backend surface.** Mass delete reuses the existing hardened `delete_session` command (path validation, symlink guard, trash-with-fallback, Codex DB cleanup) in a concurrency-limited loop with partial-failure reporting. A batch Rust command would just loop internally anyway, and this keeps Tauri ↔ Axum command parity untouched.
- New `sessionSelectionSlice` mirrors the existing `captureModeSlice` (message multi-select) — same mode / anchor / range-click shape.
- Selection is **scoped to the open project** and resets on project switch (`selectProject` / `clearProjectSelection`).
- Read-only WebUI mode hides Delete but still allows Copy IDs.
- Works in both the plain and virtualized (`react-window`) render paths.

## i18n

Added `session.selection.*` keys to all 5 locales (en, ko, ja, zh-CN, zh-TW) and regenerated `types.generated.ts`. `pnpm run i18n:validate` passes (1838 keys, no duplicates).

## Testing

- New unit tests for `sessionSelectionSlice` (toggle mode, plain/cmd/shift click, range, select all, clear).
- `pnpm tsc --build` ✓
- `pnpm vitest run` ✓ (876 passing)
- `pnpm lint` ✓
- `pnpm run i18n:validate` ✓
- `pnpm build` (vite production build) ✓

## Notes

- The Select toggle lives in the session-list controls, which only render at **≥3 sessions** (the pre-existing threshold shared with search/sort), so multi-select isn't offered for a 1–2 session project. Happy to lower that if preferred.
- Export-selected was intentionally left out of this PR; the action bar is structured so it can be added later.

## Summary by CodeRabbit

* **New Features**
  * Added multi-session selection mode with checkboxes, Shift/Cmd/Ctrl click selection, and inclusive range selection.
  * Added a sticky session selection bar with select all/clear, copy selected session IDs, and bulk delete.
  * Added a multi-delete confirmation dialog with up to 5 preview names, skipped-session messaging, and full/partial success feedback.
* **Bug Fixes**
  * Automatically exits selection mode when switching or clearing projects.
  * Selection mode disables right-click context menu behavior and supports Escape to exit (when safe).
* **Documentation**
  * Added `session.selection.*` translations (EN/JA/KO/ZH-CN/ZH-TW).
* **Tests**
  * Added unit tests for selection logic and updated list tests to stub the selection bar.


jhlee0409 (changes requested):
Thanks for this — nice design (Finder-style modifiers, scoped-to-project selection, concurrency-limited delete reusing the existing hardened `delete_session` command instead of adding a new backend surface). The explanation in the PR body of why you deliberately didn't add a batch backend command is exactly right for this codebase.

Two things need fixing before merge, both flagged by CodeRabbit and consistent with this repo's error-handling convention (see the existing single-delete flow in `useSessionEditing.ts`, which is a good reference):

1. `SessionSelectionBar.handleConfirmDelete` — wrap the `await deleteSessions(...)` call in try/catch so an unexpected store-mutation error surfaces a toast instead of becoming an unhandled rejection.
2. `useSessionBatchActions.deleteSessions` — add a `catch` around the `try { … } finally { setIsDeleting(false) }` block (the per-item failures are already handled fine; it's the outer store-mutation code that's unguarded). You can reuse the existing `session.deleteError` key — it's already in all 5 locales, so no new translation work needed.

Also worth a follow-up: the new `role="checkbox"` row indicator has no `tabIndex`/`onKeyDown`, so it's not operable from the keyboard even though it's announced as a checkbox to screen readers. Doesn't need to block this PR if you'd rather ship the mouse-driven flow now and follow up, but flagging since it's a new interactive control.

Nice-to-have: `useSessionBatchActions` (concurrency + partial-failure branches) has no unit test, unlike the well-tested `sessionSelectionSlice`. Given this is delete logic, a test for "some deletes fail" and "currently-open session gets deselected" would be reassuring.

Everything else — confirmation dialog contents, WebUI read-only gating, provider-capability filtering, scoping/reset on project switch — looks correct and I didn't find any Tauri/Axum parity or path-validation issues (no new backend command was added).

jhlee0409:
To get this landed I pushed the two review fixes directly to your branch (thanks for the solid foundation — the design made them trivial):

1. `useSessionBatchActions.deleteSessions` — added the outer `catch` so unexpected store-mutation errors toast via the existing `session.deleteError` key instead of rejecting unhandled.
2. `SessionSelectionBar.handleConfirmDelete` — wrapped in try/catch/finally so the confirm dialog always closes.

Also merged latest develop and fixed an integration regression the merge surfaced: `selectProject`'s new `exitSessionSelectionMode()` call needed a stub in the #439 test harness (`projectSlice.scanProjects.test.ts`).

CI green (903 tests). The keyboard-accessibility note on the `role="checkbox"` row stays open as a follow-up — feel free to pick it up. Merging — thanks for the contribution!

PR #476, opened 2026-07-16
fix(rename): allow renaming sessions in registered custom Claude directories

BenCello:
## Summary

CCHV supports **custom Claude configuration directories** (Settings → custom directories, plus `CLAUDE_CONFIG_DIR`). Their sessions are scanned and displayed correctly — but **"Rename in Claude Code" rejects them**:

> `Permission denied: File path must be within ~/.claude directory`

`validate_claude_path` (`rename.rs`) performs its security checks correctly, then finishes with a hardcoded root:

```rust
let claude_dir = home_dir.join(".claude");
if !canonical_path.starts_with(&claude_dir) { /* reject */ }
```

So a directory the app itself scans is not renameable. Two things suggest this is an oversight rather than deliberate policy:

1. **`delete.rs` has no root restriction at all** — session *deletion* (more destructive than rename) already works on custom directories. The asymmetry is inconsistent.
2. The check predates the custom-directory feature.

`CLAUDE_CONFIG_DIR` users hit the same wall: `projectSlice.ts` feeds `detect_claude_config_dir` into the scan, but `providers::claude::get_base_path()` always returns `~/.claude`.

## Approach

Resolve the allowed roots rather than hardcoding one:

1. `~/.claude` — always
2. each custom directory from the **persisted user settings**
3. `CLAUDE_CONFIG_DIR` when set

(2) and (3) are admitted **only after passing the existing `validate_custom_claude_path()`**, which requires an absolute, non-symlinked base with a real `projects/` subdirectory.

**Security is preserved, not relaxed:**
- Every existing check is untouched: absolute path, symlink blocking on *all* path components, `^[A-Za-z0-9_-]+$` filename regex, canonicalization.
- The allowlist comes from the **trusted persisted store**, re-validated by CCHV's own validator — **never from a caller-supplied parameter** — so it cannot be used to widen access to arbitrary paths.
- Unregistered directories and symlinked bases remain rejected (covered by tests).

No command-signature or frontend change: a `pub(crate) get_user_data_path()` helper already exists, so the validator reads settings directly.

## Known limitation (out of scope for this PR)

This PR only fixes **where** CCHV is allowed to write the rename — not the event format. CCHV's native rename appends the legacy `system/local_command` `Session renamed to: …` event. Claude Code 2.x derives its `claude -r` resume-picker title from a newer **`custom-title`** event, which CCHV does not yet write. So after renaming a session in CCHV, **the new name shows correctly in CCHV but will not appear in `claude -r`** — the picker still shows the last `custom-title` set by the CLI.

This affects **all** CCHV renames (every directory, not just custom ones) and is a pre-existing format drift, independent of this change. Worth a separate follow-up to also emit a `custom-title` event (including how reset/un-rename should clear it). Flagging here for transparency.

## Test plan

- [x] New tests: registered custom dir accepted; unregistered dir rejected; dir without `projects/` doesn't widen the allowlist; symlinked custom base rejected; `~/.claude` always allowed.
- [x] Validation split into a pure `validate_claude_path_with_roots(path, roots)` so root resolution is testable without touching real user settings.
- [x] `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test -- --test-threads=1` — 847 passing (the lone `get_git_log` failure is a local-env git-author override, green in CI).
- [x] Verified in a locally built `.app`: renaming a session in a registered custom Claude directory now succeeds (previously blocked). A normal `~/.claude` session still renames fine.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session renaming now works with valid registered custom Claude directories.
  * Improved security checks prevent renaming files outside approved Claude directories, including through path traversal or symlinks.
  * Invalid or incomplete custom directory configurations are safely ignored.
  * The default Claude directory remains supported.

### DIFF
--- a/src-tauri/src/commands/session/rename.rs
+++ b/src-tauri/src/commands/session/rename.rs
@@ -11,7 +11,7 @@ use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value};
 use std::fs::{self, File};
 use std::io::{BufRead, BufReader, Write};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::time::{SystemTime, UNIX_EPOCH};
 use tauri::command;
 use uuid::Uuid;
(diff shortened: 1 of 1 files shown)


jhlee0409:
Superseded by #478, which already contains this PR's commit (04e21161) plus the symlink-depth fix and CodeRabbit hardening. Those changes are now on main. Closing as redundant.

PR #480, opened 2026-07-17
fix(cursor-agent): clean XML wrappers, suppress [REDACTED], surface tool/shell output (#472)

### LINKED ISSUE #472: UI issues for Cursor `agent` CLI
Hello, recently I use the `agent` CLI from cursor, as an alternative to Claude Code or Open Code and the CCHV doesn't render the history very well. For example

if I type `/my-skill do something ..` in the CLI
in CCHV, I got very verbose user input and tool message

Image

There are a lot of [REDACTED] in the history

Image

Also there is no shell output, only input

Even the user message is shown in XML

Image

AdarshJ173:
## Summary

Closes #472.

Reported: when using Cursor's `agent` CLI, CCHV displays (1) raw XML tags and verbose context dumps in user messages, (2) meaningless `[REDACTED]` entries in the conversation list, and (3) completely missing shell / tool output.

All three failures live in `src-tauri/src/providers/cursor_agent.rs` — the content-block normalisation layer. No frontend changes are needed.

---

### 1. Raw XML leaking into user messages

Cursor agent wraps every user query in an XML envelope and appends a large context blob:

```
<user_query>do something</user_query><context>...hundreds of lines...</context>
```

The old parser passed the full `text` block through unchanged.

**Fix:** New `clean_user_text()` function:
- Extracts only the text inside `<user_query>…</user_query>`
- Drops everything after `</user_query>` (the context blob)
- Strips any residual `<tag>` markup with a simple char-level scanner (`strip_xml_tags()`)
- Applied only to `user` role blocks so assistant prose is unaffected

### 2. `[REDACTED]` noise polluting the conversation list

Cursor redacts sensitive file contents with either:
- `{"type":"redacted", "data":"[REDACTED]"}` — explicit redaction type
- `{"type":"text", "text":"[REDACTED]"}` — literal string sentinel

Messages whose **every** block is a redaction placeholder now pass through `is_only_redacted()` and are silently skipped in both `parse_transcript` and `search`. Messages with at least one non-redacted block are still shown; only the placeholder strings are omitted from the rendered text.

### 3. Missing shell / tool output

The old `extract_text()` helper only collected `text`-type blocks. Cursor stores command results in:
- `{"type":"tool_result", "content":[…]}` — nested content array
- `{"type":"command_output", "output":"…"}` — shell stdout/stderr

**Fix:** New `extract_text_all_blocks()` / `collect_block_text()` pair recurses through all block types. Output is prefixed with `[tool output]` or `[shell]` so users can see at a glance which parts came from tools. The session-title path retains the narrow `extract_title_text()` helper so tool noise does not pollute session summaries.

### Content emitted to the frontend

To avoid any frontend changes, the normalised content is emitted as a single-element JSON array:

```json
[{"type": "text", "text": "<clean combined text>"}]
```

The existing text-block renderer handles it without modification.

---

## Tests

All **4 existing** tests pass unchanged. **5 new** tests were added:

| Test | What it covers |
|---|---|
| `clean_user_text_strips_wrapper_and_context` | XML wrapper + context blob removed |
| `clean_user_text_no_wrapper_passthrough` | Non-wrapped text passed through as-is |
| `redacted_only_message_is_skipped` | All-redacted messages dropped |
| `mixed_redacted_message_is_kept` | Partial-redacted messages kept; placeholders removed |
| `tool_result_and_command_output_extracted` | Shell/tool output surfaces in rendered content |

---

## Scope

One file changed (`cursor_agent.rs`), pure Rust, no new dependencies, no API changes, no frontend changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cursor Agent transcript rendering by filtering fully redacted messages.
  * Added support for displaying tool results and command output in conversations.
  * Cleaned user messages by removing surrounding context and markup.
  * Improved session titles to use the first meaningful user message.

### DIFF
--- a/src-tauri/src/providers/cursor_agent.rs
+++ b/src-tauri/src/providers/cursor_agent.rs
@@ -11,6 +11,18 @@
 //! ```
 //! There is no per-line `id`/`timestamp`/`sessionId`; the session id is the
 //! transcript file's UUID stem and times come from the file mtime.
+//!
+//! ## Content-block normalisation (issue #472)
+//!
+//! Cursor agent injects several non-prose block types that must be handled
+//! before content reaches the viewer:
+//!
+//! | Block type | Behaviour |
+//! |---|---|
+//! | `{"type":"text","text":"<user_query>…</user_query>"}` | Strip XML wrapper; drop trailing `<context>` blob |
+//! | `{"type":"redacted","data":"[REDACTED]"}` | Skip message when *all* blocks are redacted |
+//! | `{"type":"tool_result","content":[…]}` | Recurse into nested content array |
+//! | `{"type":"command_output","output":"…"}` | Surface shell stdout/stderr |
 
 use super::ProviderInfo;
 use crate::models::{ClaudeMessage, ClaudeProject, ClaudeSession};
@@ -27,6 +39,10 @@ const PROVIDER_ID: &str = "cursor-agent";
 
 /// Max characters of the first user prompt used as a session title.
 const SUMMARY_MAX_CHARS: usize = 80;
+
+// ============================================================================
+// Public API
+// ============================================================================
 
 /// Detect a Cursor Agent CLI installation.
 pub fn detect() -> Option<ProviderInfo> {
@@ -51,8 +67,6 @@ pub fn get_base_path() -> Option<String> {
 }
 
 /// True if at least one `*/agent-transcripts/**/*.jsonl` exists under `base`.
-/// Keeps the provider hidden when `~/.cursor/projects` holds only non-chat data
-/// (terminals/mcps/plans).
 fn has_any_transcript(base: &Path) -> bool {
     project_dirs(base).iter().any(|p| {
         let transcripts = p.join("agent-transcripts");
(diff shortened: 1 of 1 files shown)


jhlee0409:
CI blocker: `Lint & Format` fails on `cargo fmt --check`. rustfmt reports diffs in `src-tauri/src/providers/cursor_agent.rs` at lines 201, 756, 807, 826, 839. Please run `cargo fmt` in `src-tauri/` and push — the rest of the change looks good and will be merged once formatting is clean. (Maintainer can't push the fix here since 'Allow edits by maintainers' is off.)

jhlee0409:
Thanks for this — the provider logic is solid. I've adopted it onto `develop` in #482 (you're credited as co-author) with three small fixes that were blocking CI here: rustfmt, a couple of clippy lints (`let_and_return`, `doc_markdown`), and the `tool_result_and_command_output_extracted` test input (it was pretty-printed JSON, but `parse_transcript` reads JSONL one object per line, so it parsed to zero messages — the extraction code itself was correct). Since this branch has maintainer edits disabled I couldn't push the fixups here. Closing in favor of #482, which closes #472.

PR #481, opened 2026-07-18
fix(search): reveal & highlight in-session matches inside tool cards (#429)

### LINKED ISSUE #429: [Problem] - App cannot find all relevant information from chat history
I have tested the release version 1.17.1 on Windows.
I opened the application and it nicely found all chats.

I tried to find text regarding reflectance in my conversation and I saw that the search feature works non-intuitively at least.
I would expect that the search input would search through all pieces of conversation, including questions from the model.
Here is an image that shows the problem:

Image

Please, update the program so that the search feature inside a "View Message" mode, would go through all relevant sections of conversation, would find the matching string and when pressed the up/down occurrence buttons, it would navigate the view to the occurrence.

Even better would be to render the question tool similarly as the terminal claude code application draws it - a set of questions, options for the questions, and the user selected option.
Thank you.

jhlee0409:
## What & why

In-session search already indexes `tool_use.input` content (#437/#458), but matches **inside collapsed tool cards were effectively invisible**, which is what @MostEmpire reported still happening on v1.21.0:

- A match inside an `AskUserQuestion` card highlighted the message card, but the body stayed **collapsed** — you couldn't see where the match was.
- Even when expanded, the input rendered as raw JSON with **no in-text highlight**.
- Navigating between occurrences **within one long message double-scrolled** (jumped to the message top, then down to the occurrence).

Closes #429

## Changes

- **Shared `Renderer`**: new `autoExpand` prop — a collapsed card opens when a search match is inside it (consistent with the existing `CommandRenderer`/`ThinkingRenderer` pattern).
- **Plumb search context** through the unified tool-execution path: `ClaudeContentArrayRenderer → UnifiedToolExecutionRenderer → DefaultCard / ResultBlock → ToolExecutionResultRouter` (previously dropped entirely).
- **`DefaultCard`**: auto-expand, open the `<details>`, and highlight the input JSON when the term is inside it.
- **`useScrollNavigation`**: scroll straight to the active highlight when it is already rendered (within-message navigation); only fall back to `virtualizer.scrollToIndex` to bring off-screen messages into view — removes the double-jump.
- **New `AskUserQuestionCard`**: renders questions, headers, and options (label + description, single/multi-select) CLI-style with search highlighting; **defensively falls back** to the generic JSON card on unrecognized shapes (no crash on schema drift).

No new i18n strings (renders the data itself + icons), so no locale sync needed.

## Type

- [x] Bug fix
- [x] New feature (AskUserQuestion renderer)

## Checklist

- [x] `tsc --noEmit` clean
- [x] `eslint` clean
- [x] `vitest run` — full suite green (923 tests)
- [x] Added tests: `DefaultCard.search`, `useScrollNavigation.search`, `AskUserQuestionCard`

## Summary by CodeRabbit

* **New Features**
  * Added support for an “Ask User Question” tool card, including question/option details, selection type, and optional descriptions.
  * Search now auto-expands matching tool cards and highlights matched text within inputs and results.

* **Bug Fixes**
  * Improved search scrolling to avoid duplicate scrolling and to reliably target the correct highlighted match, with safe fallbacks when content isn’t yet rendered.

* **Tests**
  * Added/expanded test coverage for the Ask User Question card, in-session search highlighting/auto-expand behavior, and search navigation scrolling behavior.

### DIFF
--- a/src/components/MessageViewer/hooks/useScrollNavigation.ts
+++ b/src/components/MessageViewer/hooks/useScrollNavigation.ts
@@ -158,58 +158,52 @@ export const useScrollNavigation = ({
   const scrollToHighlight = useCallback((matchUuid: string | null) => {
     if (!matchUuid) return;
 
-    // Use virtualizer if available
+    const viewport = getScrollViewport();
+
+    // Prefer the active highlight if it is already rendered. This is the
+    // common case when moving between occurrences WITHIN one (long) message:
+    // scroll straight to the occurrence instead of first jumping the whole
+    // message into view, which caused a visible double-scroll. (#429)
+    const activeHighlight = viewport?.querySelector(
+      '[data-search-highlight="current"]'
+    );
+    if (activeHighlight) {
+      activeHighlight.scrollIntoView({ behavior: "smooth", block: "center" });
+      return;
+    }
+
+    // Target message isn't rendered yet (virtualized off-screen) — bring it
+    // in via the virtualizer, then scroll to the highlight once it mounts.
     if (virtualizer && getScrollIndex) {
       const index = getScrollIndex(matchUuid);
       if (index !== null) {
         virtualizer.scrollToIndex(index, { align: "center" });
-        // After virtualizer scrolls, try to find the highlight element
         setTimeout(() => {
-          const viewport = getScrollViewport();
-          if (!viewport) return;
-          const highlightElement = viewport.querySelector(
+          const vp = getScrollViewport();
+          if (!vp) return;
+          const highlightElement = vp.querySelector(
             '[data-search-highlight="current"]'
           );
           if (highlightElement) {
-            highlightElement.scrollIntoView({
-              behavior: "smooth",
-              block: "center",
-            });
+            highlightElement.scrollIntoView({ behavior: "smooth", block: "center" });
+            return;
           }
+          // No highlight in the now-rendered message (e.g. match outside the
+          // visible text window) — center the message as a fallback.
+          const messageElement = vp.querySelector(
+            `[data-message-uuid="${matchUuid}"]`
+          );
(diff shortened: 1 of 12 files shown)


jhlee0409:
## Self-review + independent review pass

Ran two fresh-context critics (correctness reviewer + change-verifier) against this diff.

**Verdict**: complete & correct. Wiring verified end-to-end (search props flow ClaudeContentArrayRenderer → UnifiedToolExecutionRenderer → DefaultCard/AskUserQuestionCard → ResultBlock → ToolExecutionResultRouter); no missed callsites; new `Renderer.autoExpand` and `ResultBlock` props are optional so the other 9 tool cards compile unchanged.

Addressed two low-severity findings in `4be138e`:
- **autoExpand side effect**: it wrote through `setIsOpen` into the persisted expand/WYSIWYG-capture registry, so a search permanently flipped a message's stored expand state. Now applied as an ephemeral `isOpen || autoExpand` override — reverts when the term clears, no registry mutation.
- **Scroll scoping**: the "already-rendered highlight" lookup wasn't scoped to the target message, so a `current` mark on another message (e.g. a deep-link target rendered alongside search) could hijack the scroll. Now scoped to `[data-message-uuid="<matchUuid>"]`. Added a regression test.

Left as documented residual (pre-existing convention, cosmetic): a message that is the active match renders one `data-search-highlight="current"` per text fragment (same as CommandRenderer / ClaudeContentArrayRenderer); scroll still lands on the correct message. A precise fix needs cross-fragment match indexing — out of scope for #429.

**Verification**: `tsc -p tsconfig.app.json --noEmit` clean, `eslint` clean, full `vitest run` **924 passing**. (Note: root `tsc --noEmit` is a no-op here — root tsconfig `files:[]` — use the app project / `tsc -b`.)

PR #491, opened 2026-08-04
feat(sessions): merge cross-file Claude Code session continuations

helmet-png:
## What this does

Claude Code sometimes auto-continues a conversation that ran out of context by starting a brand-new session file rather than compacting in place: the new file's `compact_boundary` system event carries a `logicalParentUuid` that points at the last message of the OLD file, not at anything in the new one. Message loading and the session list only ever looked at one file at a time, so the earlier half of the conversation was invisible in the viewer even though it's still on disk.

This is a follow-up to #238. That issue diagnosed and fixed the same-file in-place compaction case ("compaction no longer deletes history from the JSONL, and the History Viewer already shows the full conversation"), but the cross-file continuation case — a brand new session file, not an in-place `compact_boundary` — wasn't covered, and the maintainer's comment on that issue explicitly welcomed a PR for this:

> A thin in-app integration ... Render compact boundaries with the upstream snapshot context ... We are not planning to build a parallel pre-compaction backup ourselves ... A PR that adds a thin integration along the lines above is welcome.

Two smaller, unrelated improvements found while using the app are included as separate commits (happy to split into their own PRs if preferred):
- **KaTeX math rendering** — assistant messages already rendered as Markdown, but `$inline$`/`$$block$$` LaTeX was left as plain text.
- **Narrower minimum window width** — 900px sat close to two-thirds of a common laptop screen, which made the window unusable for side-by-side/split-screen use even with both panels collapsed. Lowered to 380px.

### Cross-file session continuation
- **`chain.rs`** (new): detects a session file's dangling `compact_boundary` (a `logicalParentUuid` absent from the file's own messages), finds the sibling file that actually contains it, and walks backward through the whole chain (capped at 50 hops, cached per process lifetime). Purely a display-time join — nothing on disk is read for anything but detection, and nothing is written.
- **`load.rs`**: `load_session_messages`, `load_session_messages_paginated`, and `get_session_message_count` now resolve the chain first and operate across every file in it, oldest to newest. Sessions with no chain (the common case) hit the exact same single-file code path as before — no behavior or performance change there.
- **Session list**: now hides files that are a non-leaf link in some other session's resolved chain, so a continuation shows as one entry (the leaf, which renders the full merged history) instead of one per file. As a chain grows another hop, the previous leaf becomes superseded in turn and drops out on the next refresh.
- **`watcherSlice.ts`**: wires the existing (previously unused) `triggerProjectRefresh` into the file-watcher's debounced project-update path, so a newly created session file — including a new continuation — actually refreshes the visible session list instead of only updating an internal timestamp nothing read.

### KaTeX math rendering
- Added `remark-math` + `rehype-katex` to the shared `Markdown` component and the assistant-message renderer in `MessageContentDisplay`.

### Narrower minimum window width
- `minWidth` 900 → 380, `minHeight` 600 → 400 in `tauri.conf.json`, with the corresponding config test updated.

## Testing

- `cargo test` (single-threaded): all passing tests still pass, 3 new tests added for chain resolution and superseded-path detection.
- `cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`: clean.
- `pnpm tsc --build .`, `pnpm lint`, `pnpm exec vitest run`: clean (938 tests passing).
- Manually verified against a real multi-hundred-message session split across two files by a Claude Code auto-continuation: the chain resolves correctly (`[older.jsonl, newer.jsonl]`), `total_count` goes from 431 (single file) to 2181 (merged), the first message in the merged view is genuinely the oldest file's first message (not the "This session is being continued..." placeholder), and the session list correctly collapses the pair to one entry.
- Window minWidth verified by resizing the built app to 400×500 via `SetWindowPos` — succeeds (the old 900px floor would have rejected it).

## Test plan

- [ ] Open a project containing a session that Claude Code split via auto-continuation (two `.jsonl` files linked by `logicalParentUuid`) and confirm only the newest one is listed
- [ ] Open that session and confirm the full pre-continuation history renders, in order, with the existing `compact_boundary` divider in between
- [ ] Confirm a normal (non-chained) session's list entry, message count, and pagination behavior are unchanged
- [ ] Create a new session file that continues from an already-chained session and confirm the sidebar picks it up without a manual project switch
- [ ] Send a message containing `$E = mc^2$` and confirm it renders as a formula, not literal text
- [ ] Resize the window and confirm it can shrink well below the old 900px floor

## Release Notes

* **New Features**
  * Conversation continuations are linked and displayed chronologically across session files.
  * Superseded sessions are automatically hidden from listings.
  * Session lists refresh automatically when project files change.
  * Markdown content now supports LaTeX mathematical notation.
  * The app supports smaller window sizes, with reduced minimum dimensions.

* **Improvements**
  * Message counts and pagination now include content from linked conversation sessions.


jhlee0409:
Putting this PR back into draft pending scope separation. The continuation-chain implementation is now hardened and locally verified, but the branch also includes unrelated KaTeX/Markdown rendering and window-size changes. Please split those concerns before merging so the session-continuation change can be reviewed and landed independently.

jhlee0409:
The session-continuation-only implementation has been split into PR #496 and merged after full CI verification. This original draft also contains unrelated KaTeX/Markdown rendering and window-size changes, so it is being closed to avoid duplicate or mixed-scope merges. Those unrelated changes were intentionally not merged and can be proposed separately if still desired.

PR #500, opened 2026-08-16
feat(window): allow shrinking below 1/3 of a typical screen width

helmet-png:
## What this does

`minWidth` was 900px, which sits close to two-thirds of a common laptop screen width and made the window unusable for side-by-side/split-screen use even with both the sidebar and navigator panels collapsed.

Split out from #491 at the maintainer's request to keep scope separated from the session-continuation change (merged in #496).

## Changes

- `tauri.conf.json`: `minWidth` 900 → 380, `minHeight` 600 → 400. Narrow enough for a collapsed-panel, chat-only layout, while the sidebar/navigator panels keep their own (already existing) `minWidth` constraints in `App.tsx`.
- `tauriConfig.test.ts`: updated the pinned assertions and the "reasonable minimum" sanity bound (400 → 200) to match.

## Testing

- `pnpm exec vitest run src-tauri/tests/tauriConfig.test.ts`: 54/54 passing.
- `pnpm tsc --build .`, `pnpm lint`: clean.
- Manually verified on the built app: resized the window to 400×500 via `SetWindowPos` on Windows — succeeds (the old 900px floor would have rejected it).

## Test plan

- [ ] Resize the window and confirm it can shrink to roughly a third of a typical screen width
- [ ] Confirm the sidebar/navigator panels can still be collapsed and the remaining chat column doesn't overflow at the new minimum width

## Summary by CodeRabbit

* **New Features**
  * Reduced the application window’s minimum size to 380×400 pixels, allowing it to be used comfortably in smaller windows and displays.


jhlee0409 (approved):
Reviewed: the 380x400 minimum window size is consistent with the responsive breakpoint and the targeted configuration test, TypeScript, and lint checks pass.

PR #504, opened 2026-08-17
docs: fix broken star history chart

OctoBored:
The star history chart in the README is currently broken because the underlying service relies on GitHub's stargazer API, which now restricts access.

This PR updates the chart across all README versions (en, ko, ja, zh-CN, zh-TW) to use a working alternative that pulls data from a different source and requires no API token, restoring the star history display.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -510,6 +510,6 @@ claude-code-history-viewer://session-title/auth%20bug
 
 このプロジェクトが役に立ったら、スターをお願いします！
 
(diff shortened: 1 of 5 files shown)


jhlee0409:
Reviewed while triaging open PRs.

Not merging: the original `api.star-history.com` endpoint still returns HTTP 200 and a valid ~60KB SVG for this repo. The replacement host `star-history.dera.page` also works, but it is a third-party mirror and is not needed while the official chart is alive.

Happy to revisit if the official endpoint starts failing.

OctoBored:
Are you a bot?? The svg star-history.com returns is an error message

PR #507, opened 2026-08-18
feat(providers): add Kimi Code CLI provider

cgkoreyoshi:
Add Kimi Code CLI (`~/.kimi-code`) as a new provider: session browsing, global/project stats, and UI/i18n registration in all 5 languages.

## Changes
- `src-tauri/src/providers/kimi_code.rs`: new backend provider (scan projects, load sessions/messages from `~/.kimi-code/sessions`, `KIMI_CODE_HOME` override supported)
- `src-tauri/src/commands/stats.rs`: register `KimiCode` in provider detection and the shared stats dispatch helpers
- Frontend: `kimi-code` added to `ProviderId` / `PROVIDER_IDS`
- i18n: `common.provider.kimiCode` added to en / ko / ja / zh-CN / zh-TW
- Stats provider count test updated (29 → 30)

## Verification
- Branch synced with `main` @ v1.24.1 (merge commit included; conflicts resolved in favor of the new stats dispatch helpers, with `KimiCode` arms added)
- Quality gates green: `tsc --build`, vitest 1025 passed, `i18n:validate`, `cargo test -- --test-threads=1` 918 passed, clippy `-D warnings`, `cargo fmt --check`
- Verified against real data on macOS: 15 Kimi Code sessions detected; messages, thinking blocks, and terminal tool calls render correctly

## Summary by CodeRabbit

* **New Features**
  * Added support for the Kimi Code CLI provider.
  * Kimi Code projects, sessions, messages, statistics, and global search are now discoverable.
  * Added session monitoring and automatic project detection for Kimi Code activity.
  * Added the ability to resume Kimi Code sessions.
  * Added provider counts, filtering, labels, and localized names across supported languages.
* **Bug Fixes**
  * Improved handling of incomplete or malformed session events during loading and search.

### DIFF
--- a/src-tauri/src/commands/multi_provider.rs
+++ b/src-tauri/src/commands/multi_provider.rs
@@ -88,6 +88,7 @@ pub async fn scan_all_projects(
             "goose".to_string(),
             "grok".to_string(),
             "kimi".to_string(),
+            "kimi-code".to_string(),
             "forgecode".to_string(),
             "opencode".to_string(),
             "openinterpreter".to_string(),
@@ -176,6 +177,7 @@ pub async fn scan_all_projects(
         ("goose", providers::goose::scan_projects),
         ("grok", providers::grok::scan_projects),
         ("kimi", providers::kimi::scan_projects),
+        ("kimi-code", providers::kimi_code::scan_projects),
         ("forgecode", providers::forgecode::scan_projects),
         ("opencode", providers::opencode::scan_projects),
         ("openinterpreter", providers::openinterpreter::scan_projects),
@@ -371,6 +373,7 @@ pub async fn load_provider_sessions(
         "goose" => providers::goose::load_sessions(&project_path, exclude),
         "grok" => providers::grok::load_sessions(&project_path, exclude),
         "kimi" => providers::kimi::load_sessions(&project_path, exclude),
+        "kimi-code" => providers::kimi_code::load_sessions(&project_path, exclude),
         "forgecode" => providers::forgecode::load_sessions(&project_path, exclude),
         "opencode" => providers::opencode::load_sessions(&project_path, exclude),
         "openinterpreter" => providers::openinterpreter::load_sessions(&project_path, exclude),
@@ -484,6 +487,7 @@ fn load_non_claude_messages(
         "goose" => providers::goose::load_messages(session_path),
         "grok" => providers::grok::load_messages(session_path),
         "kimi" => providers::kimi::load_messages(session_path),
+        "kimi-code" => providers::kimi_code::load_messages(session_path),
         "forgecode" => providers::forgecode::load_messages(session_path),
         "opencode" => providers::opencode::load_messages(session_path),
         "openinterpreter" => providers::openinterpreter::load_messages(session_path),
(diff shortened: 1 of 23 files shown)


jhlee0409:
Cannot merge as-is.

This PR targets `develop` but the branch was synced with `main` @ v1.24.1, so the diff is 874 files / +213k (lockfile, docs, workflows, CLAUDE.md, etc.) and GitHub reports `CONFLICTING`.

Please rebase onto current `develop` and keep only the Kimi Code provider slice (`providers/kimi_code.rs`, stats dispatch, `ProviderId` / i18n). Then we can review the actual feature.

cgkoreyoshi:
Rebased onto current `develop` as requested. The diff is now limited to the Kimi Code provider slice: `providers/kimi_code.rs` + fixtures/tests, stats/multi-provider/watcher dispatch, `ProviderId`, and i18n — 23 files, +1507/−7, mergeable. The legacy `kimi` provider (`~/.kimi`) is untouched.

jhlee0409:
Thanks for this — and sorry to close it, because the registration work here is genuinely complete. I cross-checked every place an existing provider id appears (`multi_provider.rs`, `stats.rs`, `providers/mod.rs`, `session/mod.rs`, `watcher.rs`, `lib.rs`, `providers.ts`, `types.generated.ts`, `types/core/session.ts`, `ProjectTree/index.tsx`, `providers.utils.test.ts`) and `kimi-code` is present in all of them. Partial registration is the usual failure mode for a new provider and you avoided it.

I have merged #513 instead. Both PRs create `src-tauri/src/providers/kimi_code.rs`, so they could not coexist, and the decision came down to how much of the wire journal each one replays:

| wire event | #513 | this PR |
|---|---|---|
| `context.append_message`, role ≠ user | folded | dropped at `kimi_code.rs:453-455` (`continue`) |
| `context.undo` | undo anchor, cuts the whole turn | not handled |
| `context.apply_compaction` | replacement applied | not handled |
| `context.clear` | handled | not handled |

The `step.begin` / `content.part` / `tool.call` / `tool.result` / `step.end` aggregation here is correct — the `PendingResult` buffering so a result never lands before the assistant message holding its `tool_use`, and the EOF flush for unterminated steps, are both right, and the truncated-file handling is careful.

But without the undo, compaction and clear events, the rendered conversation diverges from what the user actually had: messages they undid still appear, and content repeats across a compaction boundary. For a history viewer that is the one thing that has to be right, and it is not a small patch — it needs the fold to carry state across the whole journal rather than per step.

There is also the open CodeRabbit finding about intermediate-directory symlinks on the `agents/main/wire.jsonl` path (only the final component is checked with `is_symlink`), which #513 handles component-wise in `join_without_symlinks`.

### What I would like to take from this PR

Your `src-tauri/tests/fixtures/kimi-code/` files and `tests/kimi_code_provider_test.rs`. #513's tests are self-contained with inline wire strings via `TempDir`, which is good for isolation but means the repo has no realistic wire sample committed. A follow-up PR adding your fixtures as an integration test against the merged `kimi_code.rs` would be genuinely useful, and I would review it promptly. Please open it if you are up for it.

Sorry for the duplicated effort — #510 should have been assigned before either of you started. That is on me.

PR #520, opened 2026-08-24
fix(webui): gate restore writes on the bind address, matching the read path

### LINKED ISSUE #519: security: WebUI restore_file has no directory allowlist, while read_text_file beside it does
In `src-tauri/src/server/handlers.rs`, the WebUI's read endpoint enforces a directory allowlist and the write endpoint beside it does not.

```rust
handler_json!(read_text_file, PathParam, |p: PathParam| async move {
    // WebUI: enforce directory allowlist (Tauri desktop relies on OS dialog)
    let path = PathBuf::from(&p.path);
    commands::claude_settings::is_safe_path(&path)?;
    commands::claude_settings::read_text_file(p.path).await
});

handler_json!(
    restore_file,
    RestoreFileParams,
    |p: RestoreFileParams| async move { commands::session::restore_file(p.file_path, p.content).await }
);
```

`restore_file` writes caller-supplied content to a caller-supplied path. Its command-level validation rejects null bytes, relative paths and `..` segments, and it writes atomically, but nothing constrains *where* it may write. So over `--serve` a request can overwrite any file the server process can reach, with content it chooses, while `read_text_file` one handler above refuses to read the very same path.

The asymmetry is the point. Writing is strictly more dangerous than reading, so if either endpoint should carry the allowlist it is this one. That the guard is present on the safer call and absent on the riskier one reads as an oversight rather than a decision.

### Severity

Bounded, but not negligible.

- Token auth is on by default, so this needs an authenticated caller or a leaked token.
- `--no-auth` on a non-loopback host is refused unless `--allow-unsafe-no-auth` is passed (`lib.rs:767-773`).
- The desktop app is unaffected: Tauri relies on the OS dialog for intent, which is what the comment on `read_text_file` refers to.

So the realistic exposure is an authenticated WebUI session, or a deliberately unsafe configuration, gaining arbitrary file write on the host as the user running the server.

### Not from the recent-edits work

Verified before filing: this handler is byte-identical in `v1.25.0` and does not appear in #515's diff. It ships in the current stable release.

### How it surfaced

While adding a diff preview to the restore confirmation (#515, 02d2a09), which reads the file to show what the write will overwrite. The read is refused for project files under `--serve` while the write it is previewing would be allowed - which is what made the asymmetry visible.

### Suggested fix

Apply the same guard the read path uses:

```rust
handler_json!(restore_file, RestoreFileParams, |p: RestoreFileParams| async move {
    let path = PathBuf::from(&p.file_path);
    commands::claude_settings::is_safe_path(&path)?;
    commands::session::restore_file(p.file_path, p.content).await
});
```

Worth deciding deliberately rather than pasting, because it has a visible consequence: restoring a file that lives outside the allowlist would then be refused in the WebUI, while continuing to work on the desktop. That may well be the correct trade, but it changes what Restore can do in the browser and should be a choice rather than a side effect.

Whichever way that goes, the two handlers should agree.

jprisant:
Closes #519.

## The problem

In `server/handlers.rs`, the WebUI's read endpoint enforces the export directory allowlist and the write endpoint one handler below enforces nothing:

```rust
read_text_file  → is_safe_path(&path)?;   // guarded
restore_file    → (no check)              // writes anywhere
```

`restore_file` writes caller-supplied content to a caller-supplied path. Its command-level validation rejects null bytes, relative paths and `..`, and writes atomically, but nothing constrains *where*. So over `--serve` a request could overwrite any file the server process could reach while being refused permission to **read** that same path.

Writing is strictly more dangerous than reading. The guard being present on the safer call and absent on the riskier one reads as an oversight rather than a decision.

## Why the bind address rather than a blanket allowlist

Applying `is_safe_path` unconditionally would have been one line, and it would have broken the feature for its main use.

The two deployments are genuinely different situations:

| bind | who the caller is | what restore may write |
|---|---|---|
| loopback | this machine | anywhere, as the desktop app already does via the OS dialog |
| routable | someone else | the same allowlist the read path uses |

On loopback, anything a request can reach the person at the keyboard already could, so the allowlist buys no safety and costs the ability to restore a file in your own project. On a routable address that reasoning disappears entirely.

## Shape

- `AppState.loopback_bind`, computed in `run_server` from `is_loopback_bind_host` - the same predicate the auth startup checks use, so "is this machine talking to itself" has one definition in the binary rather than two.
- The field is `false` at every construction site including the tests. A site that forgets it gets the restrictive behaviour, not the permissive one.
- The handler is stateful rather than `handler_json!`-generated, since the decision depends on state the macro does not pass.
- `restore_write_allowed` is named rather than inlined so it can be tested without standing up a router, the same reason `is_read_only_allowed_path` is.

Command-level validation still runs underneath in both cases. This only decides *where*.

## Not affected

The desktop app. These are `WebUI` handlers; Tauri relies on the OS dialog for intent, which is what the existing comment on `read_text_file` refers to.

Existing protections are unchanged and still apply first: token auth is on by default, `--no-auth` on a non-loopback host is refused without `--allow-unsafe-no-auth`, and `--read-only` already rejects `/restore_file` as a mutating path.

## Tests

`test_restore_write_is_unrestricted_only_on_loopback` pins both directions: a path no allowlist would cover is permitted on loopback and refused otherwise. Full server suite passes (27), `cargo fmt --check` and `cargo clippy --all-targets --all-features -D warnings` clean.

## How it surfaced

While adding a diff preview to the restore confirmation in #515, which reads the file to show what the write will overwrite. The read was refused for a path the write would have accepted, which is what made the asymmetry visible.

## Note on the second commit

`c668db9` is a cherry-pick of `d760c6d`, a Windows-only clippy false positive in `rename.rs` that blocks committing any Rust change on Windows. It is also on #515. Whichever lands first, the other carries an identical hunk. It arguably belongs here rather than there, since this is the Rust-side PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened file-restore security by allowing unrestricted writes only for authenticated access on loopback connections.
  * Servers on routable addresses and unauthenticated loopback connections now enforce the export allowlist.
  * Added safeguards to reject disallowed restore paths before any files are written.
  * Improved monitoring of Kimi Code session files when the sessions directory exists.

### DIFF
--- a/src-tauri/src/commands/session/rename.rs
+++ b/src-tauri/src/commands/session/rename.rs
@@ -373,6 +373,14 @@ fn strip_windows_extended_prefix(path: &Path) -> PathBuf {
 /// Normalize only the representation used for the allowlist boundary check.
 /// Windows paths are case-insensitive for normal filesystem access, while
 /// Unix paths retain their case-sensitive semantics.
+///
+/// The `return` in the Windows arm reads as needless to clippy, which only sees
+/// one arm at a time: on Windows the `cfg(not(windows))` tail is compiled away,
+/// so the block looks like it could just be the tail expression. Keeping the
+/// explicit `return` keeps the two arms symmetrical and independent of which
+/// one happens to come last. The lint only fires when compiling for Windows,
+/// which is why CI does not see it.
+#[allow(clippy::needless_return)]
 fn normalize_path_for_comparison(path: &Path) -> PathBuf {
     let stripped = strip_windows_extended_prefix(path);
 
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -460,6 +460,10 @@ fn run_server(args: &[String]) {
         start_time: std::time::Instant::now(),
         auth: resolved_auth.auth.clone(),
         read_only,
+        // Computed from the resolved bind host, the same predicate the auth
+        // startup checks use, so "is this machine talking to itself" has one
+        // definition in this binary rather than two.
+        loopback_bind: is_loopback_bind_host(&host),
         event_tx,
     });
 
--- a/src-tauri/src/server/handlers.rs
+++ b/src-tauri/src/server/handlers.rs
@@ -564,11 +564,36 @@ handler_json!(
 ///
 /// Named rather than inlined so the decision can be tested without standing up
 /// a router, the same reason `is_read_only_allowed_path` is its own function.
-pub(crate) fn restore_write_allowed(loopback_bind: bool, path: &Path) -> Result<(), String> {
-    if loopback_bind {
+///
+/// `unrestricted` is deliberately a decided boolean rather than the raw bind
+/// flag, so the one place that decides is the call site and this stays a pure
+/// predicate.
+pub(crate) fn restore_write_allowed(unrestricted: bool, path: &Path) -> Result<(), String> {
+    if unrestricted {
         return Ok(());
     }
     commands::claude_settings::is_safe_path(path)
+}
+
+/// Whether this server may write outside the export allowlist.
+///
+/// Loopback alone is not enough. "The caller is this machine" is only the same
+/// statement as "the caller is the person at the keyboard" while something
+/// proves the request came from this app: with `--no-auth` the router installs
+/// `allow_origin(Any)` (see `build_router`), so any page the user has open can
+/// preflight and then `POST /api/restore_file` at `127.0.0.1` with a path and a
+/// body of its choosing. `~/.zshrc`, or a hook in `~/.claude/settings.json`,
+/// makes that code execution.
+///
+/// `--no-auth` is also permitted on a loopback host without
+/// `--allow-unsafe-no-auth`, so that combination is not exotic — it is the one
+/// in the project's own `--serve` instructions.
+///
+/// Requiring both keeps the case the carve-out exists for (the desktop-like
+/// local session, authenticated, restoring a file in your own project) and
+/// drops the one it never meant to cover.
+pub(crate) fn restore_write_is_unrestricted(state: &AppState) -> bool {
+    state.loopback_bind && state.auth.is_enabled()
 }
 
 /// Write a file back to disk from a recorded edit.
@@ -582,12 +607,14 @@ pub(crate) fn restore_write_allowed(loopback_bind: bool, path: &Path) -> Result<
 /// being refused permission to *read* that same path. Writing is strictly the
 /// more dangerous of the two, so the guard belonged here first.
 ///
-/// The gate is the bind address rather than a blanket allowlist. On loopback the
-/// caller is this machine, so restoring a file anywhere is exactly what the
-/// desktop app already does through the OS dialog, and forcing the allowlist
-/// there would break restoring a file in your own project for no safety gained.
-/// Bound to a routable address the caller is someone else, and the same
-/// allowlist the read path uses applies.
+/// The gate is the bind address plus authentication rather than a blanket
+/// allowlist. On an authenticated loopback server the caller is this app on
+/// this machine, so restoring a file anywhere is exactly what the desktop app
+/// already does through the OS dialog, and forcing the allowlist there would
+/// break restoring a file in your own project for no safety gained. Bound to a
+/// routable address, or with auth off, the caller is someone else — possibly a
+/// web page — and the same allowlist the read path uses applies. See
+/// `restore_write_is_unrestricted`.
 ///
 /// Command-level validation (absolute path, no null bytes, no `..`, atomic
 /// write) still runs underneath in both cases; this only decides *where*.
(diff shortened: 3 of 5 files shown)


jhlee0409 (changes requested):
Verified the claims in the description against the tree; the factual ones hold. `/restore_file` really is on `READ_ONLY_MUTATING_API_PATHS` (`server/mod.rs:123`), every `AppState` construction site really does default `loopback_bind: false`, and `is_loopback_bind_host` behaves correctly for `0.0.0.0`, `::`, `::1`, `127.0.0.2` and `localhost`.

The asymmetry you found is real and worth closing. Two things need to change before this lands.

---

### 1. The loopback carve-out is open to any web page the user visits

`--no-auth` is permitted on a loopback host without `--allow-unsafe-no-auth` (`lib.rs:767`), and with auth disabled the router installs `allow_origin(Any)` together with `allow_headers([CONTENT_TYPE, AUTHORIZATION, x-csrf-token])` (`server/mod.rs:178-187`).

So under `--serve --no-auth --host 127.0.0.1` — the exact command used to verify this PR — any site the user has open can preflight and then `POST http://127.0.0.1:PORT/api/restore_file` with an arbitrary path and arbitrary content. `~/.zshrc` or a hook in `~/.claude/settings.json` turns that into code execution.

"On loopback the caller is this machine, so the caller is the person at the keyboard" is the premise the carve-out rests on, and a browser tab is also on this machine. The PR closes the remote vector and leaves this one explicitly open.

Suggested shape, keeping the intent:

```rust
pub(crate) fn restore_write_allowed(
    unrestricted: bool,
    path: &Path,
) -> Result<(), String> {
    if unrestricted {
        return Ok(());
    }
    commands::claude_settings::is_safe_path(path)
}
```

with the call site passing `state.loopback_bind && state.auth.is_enabled()`. Auth being on is what makes "the caller is this machine" mean "the caller is *this app*"; without it the permissive CORS hands the endpoint to anyone.

### 2. `write_text_file` one screen up disagrees

`handlers.rs:600-609` gates `write_text_file` through `is_safe_path` unconditionally, loopback or not. After this PR the file contains two write endpoints under two different policies, and the reasoning in the doc comment applies equally to both.

Pick one and state it:

- if the loopback argument holds, `write_text_file` should take the same gate; or
- if it does not, `restore_file` should be gated unconditionally and the fix is one line.

Worth noting that the underlying problem the diff-preview in #515 ran into is not the presence of the guard but the *contents* of the export allowlist — it does not cover the user's project directories, so `read_text_file` refuses a path the app legitimately needs. Widening the allowlist to include scanned project roots would fix the preview and the restore path together, under one policy, and would not need a bind-address special case at all. Worth considering as the alternative to this whole approach.

---

### Smaller

**`rename.rs` clippy hunk (P2).** Unrelated to this change, and byte-identical to a hunk on #515, so whichever lands second conflicts. Please split it into its own PR against `develop`; I will take it immediately.

**Test does not pin the wiring (P3).** `test_restore_write_is_unrestricted_only_on_loopback` exercises `restore_write_allowed` in isolation. Delete the call to it from the handler and the test still passes. A request-level case through `build_router` with a state whose `loopback_bind` is false — asserting the 4xx — would pin what actually matters.

jhlee0409:
The `rename.rs` clippy hunk has landed on `develop` as #521 (cherry-picked from your `c668db9`, authorship preserved). Please rebase and drop the duplicate hunk from this branch — otherwise whichever of #515 / #520 rebases second will conflict on it.

jhlee0409:
Pushed the narrowing to this branch rather than waiting, because the hole this PR closes is live on `develop` right now — `restore_file` there has no guard at all, so blocking the PR was leaving the strictly worse state in place. Reverting my commit is fine if you disagree with the shape.

**What changed.** `restore_write_is_unrestricted(&state)` = `loopback_bind && auth.is_enabled()`, and the handler calls that instead of reading `loopback_bind` directly. `restore_write_allowed` keeps its signature but takes a decided boolean, so it stays a pure predicate and the call site owns the decision.

Authentication is what turns "this machine" into "this app". Without it, `allow_origin(Any)` means a page that merely knows the port qualifies as "this machine" under the old condition.

**Also added an endpoint-level test.** Your original exercised `restore_write_allowed` in isolation, so deleting the guard call from the handler left it green. `test_restore_endpoint_refuses_a_path_outside_the_allowlist` drives `/api/restore_file` through `build_router` and asserts both the refusal and that nothing reached the disk. Mutation-checked: unwiring the handler fails only the new test, which is the point of it.

**Gates on the merge result** (`develop` merged in, which also collapsed the `rename.rs` hunk against #521 — nothing left for you to drop there):

- `cargo test --all-features -- --test-threads=1`: 991 + 5 + 4 + 3 passed, 0 failed
- `cargo clippy --all-targets --all-features -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean

**Still open — your call, not mine to make.** `write_text_file` one screen up is gated unconditionally, so the file still holds two write endpoints under two policies. My preference is still the alternative I raised: widen the export allowlist to include scanned project roots. That gives one policy for both endpoints, and it fixes the diff preview in #515 which currently shows "unreadable" for exactly the files it is meant to preview. If you take that route this commit becomes unnecessary and can come back out.

Merging as-is once CI is green, since it is a net improvement over develop either way.

jhlee0409 (approved):
Approving the branch as it now stands, including my own hardening commit. The remaining `write_text_file` policy question is tracked separately and does not need to hold this.

PR #537, opened 2026-08-27
fix(tests): stop cargo test overwriting the real ~/.claude/settings.json

### LINKED ISSUE #536: cargo test overwrites the developer's real ~/.claude/settings.json on Windows
## Summary

`cargo test` overwrites the developer's real `~/.claude/settings.json` with
`{"theme":"dark","fontSize":14}`, destroying every other key in it.

On this machine that included `cleanupPeriodDays: 36500`, which governs how long
Claude Code keeps session transcripts. Dropping it back to the default made
Claude Code eligible to prune transcripts on its next start. The config was
destroyed twice before anyone connected it to `cargo test`, because the test
reports `ok` while doing it.

## Cause

`setup_test_env()` in `src-tauri/src/commands/claude_settings.rs` sandboxes with:

```rust
env::set_var("HOME", temp_dir.path());
```

`HOME` is a Unix convention. On Windows `dirs::home_dir()` (dirs 5.0) resolves
through the known-folder API and consults neither `HOME` nor `USERPROFILE`, so
the sandbox silently does nothing. `test_save_and_retrieve_user_settings` then
calls `save_settings("user", ...)`, which resolves to the real path and replaces
rather than merges.

The existing `test_get_user_settings_path` did not catch it: it asserts only that
the path contains `.claude` and ends with `settings.json`, both true of the real
file.

## Reproduction

On Windows, with a populated `~/.claude/settings.json`:

```
cargo test --lib claude_settings::tests::test_save_and_retrieve_user_settings -- --test-threads=1
```

Observed here: 7273 bytes / 22 keys became 30 bytes / `{"theme":"dark","fontSize":14}`,
with the test reporting `ok`.

## Also affected

`src-tauri/src/commands/settings.rs` has the identical `setup_test_env()` against
`~/.claude-history-viewer/presets`. Less destructive, since that folder belongs
to this app, but the same mechanism. On this machine that folder carries the
timestamp of the same `cargo test` run.

## Masked by a compile error

The Rust suite does not currently compile on Windows at all: two symlink tests in
`src-tauri/src/providers/kimi_code.rs` import `std::os::unix::fs` without a
`#[cfg(unix)]` gate, so `std::os::unix` fails to resolve and the whole lib test
binary fails to build. That is why the wipe was hard to attribute. Fixing the
compile error re-arms the config wipe, so the two need fixing together.

## Environment

- Windows 11, `dirs` 5.0.1
- Reproduced on v1.26.0

jprisant:
## What & why

Closes #536

`cargo test` overwrites the developer's real `~/.claude/settings.json` with
`{"theme":"dark","fontSize":14}`, taking every other key with it. On the machine
this was found on, that included `cleanupPeriodDays: 36500`, the setting that
governs how long Claude Code keeps session transcripts. It destroyed that config
twice, and the test reported `ok` both times.

`setup_test_env()` sandboxes with `env::set_var("HOME", temp_dir)`. `HOME` is a
Unix convention: on Windows `dirs::home_dir()` resolves through the known-folder
API and consults neither `HOME` nor `USERPROFILE`, so the sandbox silently does
nothing and `save_settings` writes to the real path. It replaces rather than
merges, so the whole file goes at once.

Reproduced before fixing: **7273 bytes / 22 keys became 30 bytes.**

## The fix

Both files resolve their home through a small helper. In a normal build it is
exactly `dirs::home_dir()`. Under `cfg(test)` it reads `CCHV_TEST_HOME` and
**panics when that is unset**.

The panic is the deliberate part. Falling back to the real home would leave the
same landmine armed for the next test that forgets to sandbox itself. A panic
turns that mistake into a loud failure instead of silent data loss.

`settings.rs` had the identical defect against `~/.claude-history-viewer/presets`.
Less destructive, since that folder belongs to this app rather than to Claude
Code, but the same mechanism, and on the affected machine that folder carries the
timestamp of the same `cargo test` run.

## Two supporting fixes, both needed to see any of this on Windows

- **`kimi_code.rs` did not compile on Windows.** Two symlink tests import
  `std::os::unix::fs` with no `#[cfg(unix)]` gate, so `std::os::unix` fails to
  resolve and the **entire lib test binary** fails to build. The Rust suite could
  not run on Windows at all, which is why the config wipe went unattributed for
  days. Gated both, matching the convention already used in six other modules.
  Note this re-arms the wipe, which is why it belongs in the same PR.
- **`test_get_presets_folder` asserted a substring with a forward slash**
  (`".claude-history-viewer/presets"`), which a Windows path never contains.
  Compared by path components instead.

## Type

- [x] Bug fix
- [ ] New feature
- [ ] New provider (Claude Code / Codex / OpenCode / Kiro / Kimi / Copilot)
- [ ] Refactor / perf
- [ ] Docs

## Checklist

- [x] `pnpm exec tsc --build .` and `pnpm lint` pass (frontend untouched)
- [x] **i18n**: no user-facing strings changed

## Tests

Adds `settings_paths_stay_inside_the_sandbox`, which anchors every resolved path
inside the temp directory.

The existing `test_get_user_settings_path` asserted only that the path contains
`.claude` and ends with `settings.json`. Both were true of the developer's real
file, which is precisely why it passed while the sandbox did nothing. Checking
containment is the assertion that actually proves the sandbox holds.

## Verification

On Windows, watching the real file by hash throughout:

| | Before this PR | After |
|---|---|---|
| Lib test binary | does not compile | compiles |
| `commands::settings::` | could not run | 8 passed, 0 failed |
| `claude_settings::` (isolated) | 6 failed | 2 failed, both pre-existing and order-dependent |
| Full lib suite | could not run | 910 passed, 40 failed |
| Real `settings.json` after a full run | wiped to 30 bytes | **byte-identical, 22 keys** |

`cargo clippy --all-targets --all-features -- -D warnings` and
`cargo fmt --all -- --check` both clean.

The 40 remaining failures are pre-existing Windows environment issues, newly
visible now that the suite compiles at all. None are in the files this PR
touches, and none were introduced here.

## Note for reviewers on other platforms

On Unix `env::set_var("HOME")` already worked, so none of this was reachable
there. `HOME` is still set by `setup_test_env()` alongside the new variable,
because other helpers reached from these tests resolve through
`dirs::home_dir()` directly and behave correctly with it on Unix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved settings and preset path handling across supported platforms.
  * Fixed test path validation for Windows compatibility.
  * Prevented platform-specific tests from breaking builds on non-Unix systems.
  * Ensured test data remains contained within its temporary sandbox.

### DIFF
--- a/src-tauri/src/commands/claude_settings.rs
+++ b/src-tauri/src/commands/claude_settings.rs
@@ -39,21 +39,61 @@ pub struct AllMCPServers {
     pub local_claude_json: Option<serde_json::Value>,
 }
 
+/// The home directory these settings paths hang off.
+///
+/// In a normal build this is exactly `dirs::home_dir()`.
+///
+/// # Why this indirection exists
+///
+/// The tests below intended to sandbox themselves with
+/// `env::set_var("HOME", temp_dir)`. That is inert on Windows: `dirs::home_dir()`
+/// resolves through the known-folder API and consults neither `HOME` nor
+/// `USERPROFILE`. So `test_save_and_retrieve_user_settings` wrote
+/// `{"theme":"dark","fontSize":14}` straight over the developer's real
+/// `~/.claude/settings.json`, and because `save_settings` replaces rather than
+/// merges, every other key went with it. It destroyed a 22-key config twice
+/// before anyone connected it to `cargo test`, and the test reported `ok` each
+/// time.
+///
+/// Under `cfg(test)` this reads `CCHV_TEST_HOME` instead, and **panics if that
+/// variable is unset**. Falling back to the real home would leave the same
+/// landmine armed for the next test that forgets to sandbox itself; a panic
+/// turns that mistake into a loud failure instead of silent data loss.
+#[cfg(not(test))]
+fn settings_home_dir() -> Option<PathBuf> {
+    dirs::home_dir()
+}
+
+#[cfg(test)]
+// Always Some or a panic under cfg(test); the Option matches the real
+// signature above so callers stay identical in both builds.
+#[allow(clippy::unnecessary_wraps)]
+fn settings_home_dir() -> Option<PathBuf> {
+    match std::env::var_os("CCHV_TEST_HOME") {
+        Some(value) => Some(PathBuf::from(value)),
+        None => panic!(
+            "a test resolved a real ~/.claude path without a sandbox. \
+             Call `setup_test_env()` first: writing here would overwrite the \
+             developer's own settings.json."
+        ),
+    }
+}
+
 /// Get the user settings path (~/.claude/settings.json)
 fn get_user_settings_path() -> Result<PathBuf, String> {
-    let home = dirs::home_dir().ok_or("Could not find home directory")?;
+    let home = settings_home_dir().ok_or("Could not find home directory")?;
     Ok(home.join(".claude").join("settings.json"))
 }
 
 /// Get the user MCP settings path (~/.claude/.mcp.json)
 fn get_user_mcp_path() -> Result<PathBuf, String> {
-    let home = dirs::home_dir().ok_or("Could not find home directory")?;
+    let home = settings_home_dir().ok_or("Could not find home directory")?;
     Ok(home.join(".claude").join(".mcp.json"))
 }
 
 /// Get the main Claude config path (~/.claude.json) - the official config file
 fn get_claude_json_path() -> Result<PathBuf, String> {
-    let home = dirs::home_dir().ok_or("Could not find home directory")?;
+    let home = settings_home_dir().ok_or("Could not find home directory")?;
     Ok(home.join(".claude.json"))
 }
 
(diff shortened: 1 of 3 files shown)


jhlee0409 (approved):
Merging this. It is a data-loss bug and the diagnosis is exact. Notes from verifying it, including one thing about our CI that matters more than the patch.

## Verified

**The ungated imports are exactly two, and this gates both.** Scanned every `use std::os::unix` in `src-tauri/src` on `develop`: 15 occurrences, 3 outside a `cfg(unix)` context. Two are the `kimi_code.rs` tests you gate here. The third, `commands/session/mod.rs:148`, is already covered — its whole module is `#[cfg(all(test, feature = "webui-server", unix))]`. So the set is complete, not a sample.

**macOS run is clean and writes nothing.** Fingerprinted my real `~/.claude/settings.json` before and after `cargo test --all-features -- --test-threads=1`:

```
before  dc9057e3a062436e…  5051 bytes
after   dc9057e3a062436e…  5051 bytes
```

1022 + 6 + 5 + 4 + 3 passed, 0 failed, including the new `settings_paths_stay_inside_the_sandbox`. `~/.claude-history-viewer/presets` still carries January/April timestamps, so nothing landed there either — consistent with the Unix sandbox having always worked and the defect being Windows-only.

**The panic-instead-of-fallback choice is right.** A fallback would leave the landmine armed for the next test that forgets `setup_test_env()`; the panic converts that into a loud failure. Worth the ugliness.

## The part that is bigger than this PR

**Our CI cannot verify this fix, and could not have caught the bug.**

`rust-tests.yml` runs `ubuntu-latest` only on pull requests, and `ubuntu-latest, macos-latest` on push. **Windows is never built.** That is why the lib test binary has been failing to compile on Windows since #513 merged, and why the config wipe went unattributed.

It also means the new assertion is non-discriminating on every platform we do test: on Unix `dirs::home_dir()` reads `$HOME`, so `settings_paths_stay_inside_the_sandbox` passes with or without the `CCHV_TEST_HOME` indirection. The test is correct and worth having — it just cannot fail anywhere CI looks.

I will add `windows-latest` to the matrix. Without it the next `use std::os::unix` in a test module breaks the Windows build again and we find out the same way.

## Residual, not blocking

`#[cfg(test)]` applies to the lib's own unit tests. Integration tests under `src-tauri/tests/` link the lib built **without** `cfg(test)`, so they would resolve the real home. Nothing there touches settings or presets today — I checked — so this is latent rather than live. Worth a comment at the helper so the next person adding an integration test does not assume they are covered.

## On how this got in

#513 introduced the ungated imports and I merged it after running the Rust suite on macOS and reporting "946 tests pass" as if that cleared it. It did not clear Windows, and I did not say so. Your PR is cleaning up after my verification gap, and the description was more careful about platform scope than my merge note was.

PR #539, opened 2026-08-28
ci(rust): add a Windows build gate, and an informational Windows test run

jhlee0409:
Closes the systemic half of #536. #537 fixed the data loss; this closes the reason nobody saw it for days.

## The gap

`rust-tests.yml` ran `ubuntu-latest` on pull requests and `ubuntu-latest, macos-latest` on push. **Windows was never built.** Two consequences, both of which actually happened:

1. Two `use std::os::unix` imports landed in a `kimi_code.rs` test module without a `cfg(unix)` gate. An unresolved path in any test module fails the *whole* lib test binary, so the Rust suite had not compiled on Windows since that merged. Nothing reported it, because nothing built it.
2. The settings tests sandbox with `env::set_var("HOME", temp)`, which is inert on Windows — `dirs::home_dir()` there resolves through the known-folder API. Writes landed on the developer's real `~/.claude/settings.json`, and the suite reported `ok` while doing it.

The second was found by a human losing their config twice, not by CI.

## Why #537's own test does not cover this

`settings_paths_stay_inside_the_sandbox` cannot fail on any platform CI currently runs. On Unix `dirs::home_dir()` reads `$HOME`, so the assertion holds with or without the `CCHV_TEST_HOME` indirection. The test is correct and worth having — the matrix was the gap, and no test written in that file can close it.

## The change

```yaml
os: ${{ github.event_name == 'pull_request'
        && fromJSON('["ubuntu-latest", "windows-latest"]')
        || fromJSON('["ubuntu-latest", "macos-latest", "windows-latest"]') }}
```

**Windows is on the pull-request matrix, not only on push.** That is the deliberate part. It is the platform no contributor here runs locally, so it has to be a pre-merge gate rather than a post-merge discovery — post-merge is exactly what we had, and it took a lost config to surface. macOS stays push-only, because contributors do run it and it is the least likely to hold a surprise.

One supporting fix: the dist placeholder step gains `shell: bash`. The Windows runner defaults to PowerShell, which has no `touch` and rejects `mkdir -p`.

## Type

- [x] Refactor / perf (CI)

## Checklist

- [x] PR targets `develop`
- [x] No product code touched; frontend and Rust sources unchanged
- [x] i18n: not applicable

## What this PR is also for

I cannot run Windows locally, so **this PR's own CI run is the experiment.** It is the first time the Rust suite will have been built and run on Windows since #513.

If it comes back green, the matrix change stands as written. If it comes back red, that is not a reason to revert it — it is the backlog it was added to find, and I will triage what it turns up before merging rather than after.

## Summary by CodeRabbit

* **Tests**
  * Updated automated testing across Ubuntu, macOS, and Windows environments.
  * Pull request checks now run on Ubuntu, while push validation includes Ubuntu and macOS.
  * Added dedicated Windows build validation, with test results reported informationally.
  * Improved cross-platform reliability for workflow steps.

### DIFF
--- a/.github/workflows/rust-tests.yml
+++ b/.github/workflows/rust-tests.yml
@@ -59,24 +59,18 @@ jobs:
         run: cargo clippy --all-targets --all-features -- -D warnings
 
   # Main test job - runs in parallel with lint
+  # PR: Ubuntu only, Main: Ubuntu + macOS
   #
-  # Windows is in the matrix because it was absent, and its absence was not
-  # theoretical: two `use std::os::unix` imports landed in a test module and the
-  # entire lib test binary stopped compiling on Windows for days (#536). Nothing
-  # noticed, because nothing built it. The same blind spot let a test overwrite
-  # the developer's real `~/.claude/settings.json`, since the `HOME` sandbox those
-  # tests rely on is inert there.
-  #
-  # PR: Ubuntu + Windows. Windows is the platform no contributor runs locally, so
-  # it has to be a pre-merge gate rather than a post-merge discovery.
-  # Push: adds macOS for full coverage.
+  # Windows is NOT here. It gets its own job below, because the Windows suite
+  # does not pass yet (#540, #541) and putting it in this matrix would make
+  # every PR red for reasons unrelated to the PR.
   test:
     name: Test Suite
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        os: ${{ github.event_name == 'pull_request' && fromJSON('["ubuntu-latest", "windows-latest"]') || fromJSON('["ubuntu-latest", "macos-latest", "windows-latest"]') }}
+        os: ${{ github.event_name == 'pull_request' && fromJSON('["ubuntu-latest"]') || fromJSON('["ubuntu-latest", "macos-latest"]') }}
         rust: [stable]
 
     steps:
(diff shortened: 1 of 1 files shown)


jhlee0409:
## Result of the experiment, and a revised approach

The first push put `windows-latest` in the main matrix. That run is the first time the Rust suite has been **built and run on Windows** since #513, and it found more than expected.

**968 tests run: 928 passed, 41 failed.**

So the matrix as originally written would have reddened every unrelated PR. Revised: Windows is now its own job, split by readiness.

- **Blocking — does the test binary build?** That is the failure that hid everything else. Cheap, passes today, and it is what keeps the platform from going dark again.
- **Informational — does the suite pass?** Not yet. Reports without gating, so the backlog is visible on every PR without blocking work that has nothing to do with it.

`continue-on-error` comes off and the job folds back into the matrix once the two issues below are closed.

## What the 41 failures actually are

Two classes, and one of them is not a test problem.

### #540 — six modules still write to the real home, one of them mutates archives

`env::set_var("HOME", temp)` is inert on Windows. #537 fixed `claude_settings.rs` and `settings.rs`. **Six more modules sandbox the same way and were not covered**: `archive.rs`, `stats.rs`, `metadata.rs`, `providers/pi.rs`, `providers/antigravity.rs`, `providers/ompi.rs`. (`session/mod.rs` also uses the pattern but its module is `cfg(all(test, …, unix))`, so it never builds there.)

`archive.rs` is the serious one. `get_archives_dir()` resolves `~/.claude-history-viewer/archives` through `dirs::home_dir()`, and its test module has **nine mutating tests** — create, delete, rename, and a UUID-to-name migration. On a real machine that directory holds saved session backups.

The failure shapes are the tell. These are not logic failures, they are contamination:

```
test_load_manifest_default_when_missing   assertion failed: manifest.archives.is_empty()
test_get_archive_disk_usage_empty         left: 3   right: 0
test_get_archive_disk_usage_with_archive  left: 4   right: 1
```

"Empty" is not empty because the test is reading the developer's real archive store.

#536 overwrote a config file. This class can rename, migrate or delete the archives someone made specifically so they would not lose those sessions. **It is a continuation of the bug #537 fixed, not a separate curiosity** — the fix there was correct but covered a quarter of the surface.

### #541 — the other 35 are Unix path assumptions

Hardcoded `/tmp/project` and `/usr/lib` fixtures that are not absolute on Windows, so a validation guard fires before the behaviour under test is reached; and assertions comparing assembled paths against strings containing `/`. Test-side only, no user data involved.

## What this changes about the PR

It was proposed as CI hygiene. It turned out to be the thing that found an unfixed data-loss surface within a day of the last one, on a platform where `cargo test` is currently unsafe to run on a developer machine.

That is also the argument for the blocking half being worth having on its own: the compile gate is what makes it possible for the informational half to ever report anything.

Issue #541, opened 2026-08-28
Windows: 35 test failures from Unix path assumptions in assertions and fixtures


jhlee0409:
Found by adding Windows to the Rust matrix (#539). The suite compiles there now (after #537) and runs **968 tests: 928 pass, 41 fail.**

Six of the failures are the data-safety class tracked in #540. **This issue is the other 35**, which are test-side platform assumptions: they make the Windows suite red but do not touch user data.

## Families

**1. Unix absolute paths used as fixtures.** `/tmp/project`, `/usr/lib`, `/some/stale/Dev` are not absolute on Windows, so a path-validation guard fires before the behaviour under test is reached. The assertion then fails on the wrong error:

```
commands/metadata.rs:377      validate_project_metadata_key("/tmp/project").is_ok()
commands/project.rs:959       left: "Path must be absolute"
                              right: "Path does not exist or is not a directory"
commands/session/delete.rs:115  left: "Session path must be absolute"
                                right: "Only .jsonl session files can be deleted"
commands/claude_settings.rs:1076  result.unwrap_err().contains("'..'")
```

Fix shape: build fixtures from `TempDir` or a `cfg`-selected root rather than hardcoding `/tmp`, so the guard sees an absolute path on both platforms and the test reaches what it is actually testing.

**2. Forward-slash assertions.** A Windows path never contains `/`, so substring and equality checks on assembled paths fail:

```
kimi_provider_test.rs:17
  left:  "kimi://…\\sessions\\project-hash"
  right: "kimi://…\\sessions/project-hash"

utils.rs:855  test_extract_project_name_with_hyphens_in_segments
commands/project.rs:608  left: "-Users-jack-client-myapp"  right: "client-myapp"
```

Fix shape: compare by `Path` components, as #537 did for `test_get_presets_folder`, rather than by substring.

**3. Snapshot tests** (`models::snapshot_tests::project_snapshots`, 2 cases) — almost certainly the same separator issue baked into stored snapshots. Needs a platform-independent normalisation before comparison rather than per-platform snapshots.

## Full module breakdown

```
  6  commands::archive          → #540, not this issue
  4  commands::project
  3  utils
  3  providers::antigravity
  3  commands::stats
  3  commands::session::delete
  3  providers::codebuddy
  2  providers::pi
  2  commands::session::load
  2  models::snapshot_tests
  2  commands::claude_settings
  1  each: kimi_provider_test, commands::metadata, providers::ompi,
[...]

PR #544, opened 2026-08-28
fix(tests): sandbox the remaining modules that resolved the real home

### LINKED ISSUE #540: cargo test still writes to the real home on Windows: six modules unfixed, one of them mutates archives
Extends #536. #537 fixed two modules; **six more have the same defect, and one of them mutates the user's archived sessions.**

## The pattern

`env::set_var("HOME", temp_dir)` is inert on Windows — `dirs::home_dir()` resolves through the known-folder API and consults neither `HOME` nor `USERPROFILE`. Any test module sandboxing that way writes to the developer's real home.

Nine modules sandbox with `set_var("HOME")`:

| module | status |
|---|---|
| `commands/claude_settings.rs` | fixed in #537 |
| `commands/settings.rs` | fixed in #537 |
| `commands/archive.rs` | **unfixed — mutating** |
| `commands/stats.rs` | unfixed |
| `commands/metadata.rs` | unfixed |
| `providers/pi.rs` | unfixed |
| `providers/antigravity.rs` | unfixed |
| `providers/ompi.rs` | unfixed |
| `commands/session/mod.rs` | safe — module is `#[cfg(all(test, feature = "webui-server", unix))]`, never built on Windows |

## Why `archive.rs` is worse than the original

`get_archives_dir()` (`archive.rs:135`) resolves `~/.claude-history-viewer/archives` through `dirs::home_dir()`, and the test module contains **nine mutating tests** — create, delete, rename, and a UUID-to-name migration.

On a real machine that directory holds saved session backups:

```
~/.claude-history-viewer/archives/
  archive-manifest.json
  client-chronicles-of-seojeo-2026.-3.-6_953fdece
```

#536 overwrote a config file, which is annoying but re-creatable. This class can rename, migrate or delete **the archives a user made specifically so they would not lose those sessions.**

## How it was found

Adding `windows-latest` to the Rust matrix (#539). The suite now compiles on Windows thanks to #537 and runs 968 tests: **928 pass, 41 fail.** Six of the failures are in `archive.rs` and their shapes are the tell — they are not logic failures, they are contamination:

```
test_load_manifest_default_when_missing   assertion failed: manifest.archives.is_empty()
test_list_archives_empty                  assertion failed: manifest.archives.is_empty()
test_get_archive_disk_usage_empty         left: 3   right: 0
test_get_archive_disk_usage_with_archive  left: 4   right: 1
```

"Empty" is not empty and the counts are inflated because the tests are reading the developer's real archive store.

## Fix

Same shape as #537: an `archives_home_dir()` / `stats_home_dir()` / … helper that is `dirs::home_dir()` in a normal build and reads `CCHV_TEST_HOME` under `cfg(test)`, panicking when unset. #537 established the pattern and the reasoning for the panic; this is applying it to the remaining six.

Worth considering a single shared helper rather than one per module, since the count is now eight and the next module to grow a test will face the same choice. The trade-off is that `cfg(test)` is per-crate, not per-module, so a shared helper is fine — it just needs a home that does not create a dependency cycle.

## Note on the other 35 failures

The rest are test-side platform assumptions and are much less urgent — Unix absolute paths (`/tmp/project`, `/usr/lib`) that are not absolute on Windows, and assertions on `/` separators. They make the Windows suite red; they do not touch user data. Tracking them separately from this.

jhlee0409:
Closes #540. Extends #537, which fixed two of eight modules.

## The remaining six

`env::set_var("HOME", temp)` is inert on Windows — `dirs::home_dir()` resolves through the known-folder API and consults neither `HOME` nor `USERPROFILE`. Every test module sandboxing that way wrote to the developer's real home.

| module | before |
|---|---|
| `commands/archive.rs` | **unfixed, and mutating** |
| `commands/stats.rs` | unfixed |
| `commands/metadata.rs` | unfixed |
| `providers/pi.rs` | unfixed |
| `providers/antigravity.rs` | unfixed (reaches home via callees) |
| `providers/ompi.rs` | unfixed (reaches home via callees) |

`archive.rs` is the serious one. `get_archives_dir()` resolves `~/.claude-history-viewer/archives` and its test module has **nine mutating tests** — create, delete, rename, and a UUID-to-name migration. On a real machine that directory holds saved session backups. #536 overwrote a config file, which is re-creatable; this class can delete the archives someone made precisely so they would not lose those sessions.

## Shape

One shared `crate::utils::home_dir()` rather than a helper per module. `cfg(test)` is per-crate, so a single definition covers every caller, and eight copies of a security-relevant branch is eight chances for one to drift. It keeps #537's design otherwise: `dirs::home_dir()` normally, `CCHV_TEST_HOME` under `cfg(test)`, **panic when that is unset**.

## What the panic found immediately

**Six tests were reaching the real home with no sandbox at all** — five in `stats.rs`, one in `watcher.rs`. None of them wants a home directory; each builds its own `TempDir`, and the code under test resolves home while detecting providers. They were reading whatever happened to be in the developer's `~`. Each now points the sandbox at its own temp dir.

**`CCHV_TEST_HOME` leaks between tests** under `cargo test -- --test-threads=1`, because a `TempDir` going out of scope does not unset an environment variable. Two tests sandboxing via `EnvVarGuard::set("HOME", …)` were inheriting a stale value from an earlier test and asserting against the wrong root. They now set both, and the three provider `HomeGuard`s remove the variable on drop.

Neither of those was in the issue. Both were invisible before the panic existed.

## Type

- [x] Bug fix

## Checklist

- [x] PR targets `develop`
- [x] `cargo clippy --all-targets --all-features -- -D warnings` and `cargo fmt --all -- --check` clean
- [x] Frontend untouched
- [x] i18n: no user-facing strings

## Verification

`cargo test --all-features -- --test-threads=1`: **1022 + 6 + 5 + 4 + 3 passed, 0 failed.**

**Mutation check.** Removing archive's `CCHV_TEST_HOME` line fails 18 tests, every one of them with the actionable message rather than silently proceeding:

```
a test resolved a real home directory without a sandbox. Set CCHV_TEST_HOME
(see `setup_test_env`) before touching any path under it: writing there would
hit the developer's own files.
```

That is the guarantee: an unsandboxed archive test now cannot run.

**One mutation I deliberately did not run.** The pair that would dramatise the bug — remove the sandbox *and* make the helper fall back to the real home — would have executed nine create/delete/rename/migrate tests against my actual `~/.claude-history-viewer/archives`. That is the issue, not a demonstration of it. I confirmed my archives are intact instead (manifest still 1 entry, directory timestamps unchanged).

## Windows

The failure mode is Windows-only and I cannot run Windows locally. The informational Windows job added in #539 is the signal — the six `archive.rs` failures there were contamination from the real store (`assertion failed: manifest.archives.is_empty()`), so they should disappear with this. That job does not gate, so if some remain it reports rather than blocks.

This should also shrink #541: `stats`, `metadata`, `pi`, `antigravity` and `ompi` all appear in that list, and a share of their failures were contamination rather than path assumptions. Worth re-reading that issue against a fresh Windows run once this lands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of archive, metadata, statistics, session, and provider path handling across platforms.
  * Prevented test operations from accessing or modifying the developer’s actual home directory.
  * Strengthened test isolation, particularly on Windows.

### DIFF
--- a/src-tauri/src/commands/archive.rs
+++ b/src-tauri/src/commands/archive.rs
@@ -133,7 +133,7 @@ pub struct ExportResult {
 
 /// Returns the archives base directory path: `~/.claude-history-viewer/archives/`
 fn get_archives_dir() -> Result<PathBuf, String> {
-    let home = dirs::home_dir().ok_or("Could not find home directory")?;
+    let home = crate::utils::home_dir().ok_or("Could not find home directory")?;
     Ok(home.join(".claude-history-viewer").join("archives"))
 }
 
@@ -1313,7 +1313,7 @@ pub async fn get_expiring_sessions(
 
         // Read cleanupPeriodDays from ~/.claude/settings.json
         let cleanup_period_days: i64 = {
-            let home = dirs::home_dir().ok_or("Could not find home directory")?;
+            let home = crate::utils::home_dir().ok_or("Could not find home directory")?;
             let settings_path = home.join(".claude").join("settings.json");
             if settings_path.exists() {
                 fs::read_to_string(&settings_path)
@@ -1500,6 +1500,10 @@ mod tests {
     fn setup_test_env() -> TempDir {
         let dir = TempDir::new().unwrap();
         env::set_var("HOME", dir.path());
+        // The one that actually holds on Windows. Without it these tests ran
+        // create/delete/rename/migrate against the real
+        // `~/.claude-history-viewer/archives` (#540).
+        env::set_var("CCHV_TEST_HOME", dir.path());
         dir
     }
 
--- a/src-tauri/src/commands/metadata.rs
+++ b/src-tauri/src/commands/metadata.rs
@@ -57,7 +57,7 @@ impl Default for MetadataState {
 
 /// Get the metadata folder path (~/.claude-history-viewer)
 fn get_metadata_folder() -> Result<PathBuf, String> {
-    let home = dirs::home_dir().ok_or("Could not find home directory")?;
+    let home = crate::utils::home_dir().ok_or("Could not find home directory")?;
     Ok(home.join(".claude-history-viewer"))
 }
 
@@ -324,6 +324,9 @@ mod tests {
         let guard = TEST_ENV_MUTEX.lock().unwrap();
         let temp_dir = TempDir::new().unwrap();
         env::set_var("HOME", temp_dir.path());
+        // `HOME` is inert on Windows; this is what `crate::utils::home_dir()`
+        // reads under `cfg(test)` (#540).
+        env::set_var("CCHV_TEST_HOME", temp_dir.path());
         (guard, temp_dir)
     }
 
--- a/src-tauri/src/commands/session/rename.rs
+++ b/src-tauri/src/commands/session/rename.rs
@@ -1006,6 +1006,21 @@ pub async fn rename_opencode_session_title(
 mod tests {
     use super::*;
 
+    /// Points home resolution at an empty temp directory for the duration of a
+    /// test.
+    ///
+    /// These tests assert on rejection paths and want nothing from the home
+    /// directory, but the code under test reaches it — `configured_claude_dirs`
+    /// reads `~/.claude-history-viewer/user-data.json` and honours the custom
+    /// Claude paths registered there. That made the allowlist depend on the
+    /// developer's own machine, silently, until the sandbox in
+    /// `crate::utils::home_dir` made it say so (#540).
+    fn isolated_home() -> tempfile::TempDir {
+        let dir = tempfile::TempDir::new().expect("temp home");
+        std::env::set_var("CCHV_TEST_HOME", dir.path());
+        dir
+    }
+
     fn sample_rename_test_user(session_id: &str, uuid: &str, content: &str) -> String {
         serde_json::json!({
             "parentUuid": Value::Null,
@@ -1601,6 +1616,7 @@ mod tests {
 
     #[test]
     fn test_validate_claude_path_rejects_relative_path() {
+        let _home = isolated_home();
         let result = validate_claude_path("relative/path/file.jsonl");
         assert!(result.is_err());
         assert!(result.unwrap_err().contains("must be absolute"));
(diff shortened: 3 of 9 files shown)


jhlee0409:
## Windows result: 41 → 33 failures, and **zero in `commands::archive`**

The informational Windows job is the only place this fix is observable, and it confirms the data-safety claim directly:

| module | before | after |
|---|---|---|
| `commands::archive` | 6 | **0** |
| `providers::pi` | 2 | **0** |
| `commands::stats` | 3 | 2 |
| `commands::metadata` | 1 | 1 |
| total | 41 | 33 |

The nine mutating archive tests — create, delete, rename, migrate — no longer run against a real `~/.claude-history-viewer/archives`. That was the point of the issue.

`stats` and `metadata` keep a couple of failures each, and those are the #541 class (Unix path assumptions), not contamination.

## What CI found that my local run could not

I pushed the first commit having run `cargo test -- --test-threads=1` locally, all green. CI failed with nine more tests panicking.

The difference is the runner. `cargo test` shares one process, so `CCHV_TEST_HOME` set by an earlier test **leaks forward** and silently covers for tests that never sandbox themselves. CI uses nextest, one process per test, so there is nothing to inherit.

The nine were genuinely reaching the real home:

- five `rename.rs` validate tests, transitively — `configured_claude_dirs` reads `~/.claude-history-viewer/user-data.json` and honours the custom Claude paths registered there, so the allowlist under test depended on the developer's machine
- four `stats.rs` provider-detection tests

I installed `cargo-nextest` locally and reproduced CI exactly rather than pushing again to find out: **995 and 1040 tests, both feature sets, all green.**

## One thing deliberately reverted

Routing `rename.rs`'s own `resolve_claude_roots` through the shared helper cascaded into ten more tests. `rename.rs` reads the home but never writes there, so it carries no data-loss risk and does not belong in this issue. It is filed as #545 along with `test_validate_claude_path_valid_path`, which walks the real `~/.claude/projects` and **asserts nothing at all when it finds no file** — so the only test covering `validate_claude_path`'s accept path is vacuous on CI while appearing covered.

All checks green: Ubuntu, Lint & Format, Windows build gate.

PR #369, opened 2026-06-13
fix(session): resolve Claude resume cwd from metadata

bugsyyu:
## Summary
- Resolve Claude project `actual_path` from the JSONL `cwd` field before falling back to storage-folder decoding
- Prevent resume commands from using lossy paths derived from `.claude/projects/-...` directory names
- Add a regression test for storage names that use `-` while the real cwd contains `_`

## Root Cause
Claude project storage directory names are not a lossless representation of the original working directory. For example, a session stored under `.claude/projects/-home-cym-claude-prompt-design` can contain JSONL entries with the real cwd `/home/cym/claude_prompt_design`. The UI builds the copied resume command from `project.actual_path`; when `scan_projects` decoded the storage folder name, it produced `/home/cym/claude-prompt-design`, so the copied command failed at `cd`.

## Validation
- `cargo test --manifest-path src-tauri/Cargo.toml --features webui-server commands::project::tests::test_scan_projects`
- `cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets --all-features -- -D warnings`
- `corepack pnpm exec tsc --noEmit`
- `cargo build --manifest-path src-tauri/Cargo.toml --features webui-server`
- WebUI API smoke test against real `/root/.claude`: `scan_projects` now reports `/root/.claude/projects/-home-cym-claude-prompt-design` with `actual_path: /home/cym/claude_prompt_design`

## Summary by CodeRabbit

* **Bug Fixes**
  * Project paths now prefer workspace info embedded in session files (direct > nested), reducing incorrect storage-derived paths.

* **Improvements**
  * Project names in the UI use session metadata when available.
  * Selecting a session avoids unnecessary project switches by matching provider-aware sessions.
  * Copy/resume actions use the loaded project's actual path when available.
  * Toast-based error reporting for session selection added.

* **Tests**
  * Added coverage for JSONL-derived path precedence, session-selection behavior, and session-file parsing.

* **Chores**
  * Metadata cache version bumped to refresh session metadata.

PR #549, opened 2026-08-28
fix(windows): make project-folder decoding reachable — marker, leading dash, and root

### LINKED ISSUE #548: Windows: project-folder decoding is unreachable — hardcoded '/' marker, leading-dash assumption, and '/'-rooted decoder
Found while triaging the Windows failures in #541. Most of those are test fixtures written with Unix path literals — cosmetic. This is not: **the project-folder decoding layer cannot run at all on Windows**, and the tests only pointed at it.

Three independent defects, each on its own sufficient to disable the whole path.

## The Windows encoding, confirmed

From a real Windows user on Claude Code 2.1.31 / PowerShell — anthropics/claude-code#23090:

```
D:\OneDrive - Comp Name\Office\Comp Name\Claude Projects
→  ~/.claude/projects/D--OneDrive---Comp-Name-Office-Comp-Name-Claude-Projects
```

So on Windows `:`, `\` and spaces all collapse to `-`, and — decisively — **the encoded name has no leading dash**, because the path starts with a drive letter rather than a separator.

## Defect 1 — the marker hardcodes a forward slash

`src-tauri/src/utils.rs:278` and `:337`

```rust
const MARKER: &str = ".claude/projects/";
let marker_pos = session_storage_path.find(MARKER)?;
```

The storage path on Windows is `C:\Users\jack\.claude\projects\…`. `str::find` is plain string matching, so this never matches. Step 2 of both `decode_project_path` and `decode_project_path_verified` is unreachable.

## Defect 2 — the leading-dash assumption

`utils.rs:281` and `:340`

```rust
let stripped = encoded.strip_prefix('-')?;
```

Unix paths start with `/`, so the encoded name starts with `-`. Windows paths start with `D`, so it does not. Even with defect 1 fixed, decoding is skipped.

## Defect 3 — the decoder roots at `/` and joins with `/`

`utils.rs:390`, `:392`, `:407`

```rust
let candidate = if base_path.is_empty() { format!("/{segment}") } else { format!("{base_path}/{segment}") };
```

For `D--OneDrive-…` the first candidate is `/D`. There is no such path on Windows, so every branch fails and `decode_with_filesystem_check` returns `None` unconditionally.

There is no `cfg(windows)` and no `MAIN_SEPARATOR` anywhere in `utils.rs` — I checked; zero occurrences.

## Proof

All three are pure string logic, so they demonstrate on any host:

```
defect 1 - MARKER hardcodes '/'
   unix path finds marker : true
   win  path finds marker : false   <= decode step 2 unreachable

defect 2 - leading-dash assumption
   unix encoded strip_prefix('-'): true
   win  encoded strip_prefix('-'): false   <= decode skipped even if marker matched

defect 3 - decoder roots at '/' and joins with '/'
   candidate built for first segment of the win name: "/D"
```

## What Windows users actually get

Only strategy 1 survives — `sessions-index.json` → `originalPath`, which is a real absolute path from JSON and therefore platform-correct. When that file is missing, or its `originalPath` no longer exists on disk:

- `decode_project_path` falls through **every** branch and returns the storage path unchanged (`utils.rs:298`), so the raw `…\.claude\projects\D--OneDrive---…` folder is surfaced as the project path
- `decode_project_path_verified` returns `None`, so #369's "the containing folder is authoritative, not a stale embedded `cwd`" never applies — a project that was moved keeps showing its stale path

The two callers are the primary user-facing ones: `commands/project.rs:313` (the project list) and `commands/session/load.rs:614` (session loading).

Worth noting the encoding is *more* ambiguous on Windows than on Unix — spaces and `:` also become `-` — so the filesystem-checked decoder is needed **more** there, and it is exactly what is unavailable.

## Why this went unseen

`cli.rs` already parses Windows backslash paths, with tests (`parses_abs_windows_backslash_path_as_path_kind`), so Windows support is intended. The decode layer simply never got the same treatment, and until #539 added Windows to the matrix last week nothing ever executed it on that platform.

## On fixing it

Defect 1 alone must **not** be fixed in isolation. Once the marker matches, the heuristic at `utils.rs:288-295` starts firing on Windows names and builds `/D--OneDrive/…`-shaped garbage — strictly worse than today's "return the input unchanged". The three need to land together, with the decoder taught the drive-letter root form and the missing leading dash.

jhlee0409:
Closes #548.

Found while triaging #541's Windows failures. Most of those are test fixtures written with Unix path literals. This is not: **the project-folder decoding layer could not run at all on Windows**, and the failing tests were only pointing at it.

## The encoding, confirmed

Not guessed — taken from a real Windows install, Claude Code 2.1.31 / PowerShell, in anthropics/claude-code#23090:

```
D:\OneDrive - Comp Name\Office\Comp Name\Claude Projects
→  ~/.claude/projects/D--OneDrive---Comp-Name-Office-Comp-Name-Claude-Projects
```

`:`, `\` and spaces all collapse to `-`. And decisively: **no leading dash**, because a Windows path starts with a drive letter, not a separator. Every assumption below follows from that one difference.

## Three defects, each fatal on its own

**1. The marker hardcoded a forward slash** — `utils.rs:278`, `:337`

```rust
const MARKER: &str = ".claude/projects/";
```

`str::find` on `C:\Users\jack\.claude\projects\…` never matches. Step 2 of both `decode_project_path` and `decode_project_path_verified` was unreachable.

**2. The leading-dash assumption** — `utils.rs:281`, `:340`

```rust
let stripped = encoded.strip_prefix('-')?;
```

`D--OneDrive-…` has no leading dash, so even past the marker the name was dropped.

**3. The decoder rooted at `/` and joined with `/`** — `utils.rs:390`, `:392`, `:407`

```rust
let candidate = if base_path.is_empty() { format!("/{segment}") } else { format!("{base_path}/{segment}") };
```

First candidate for `D--…` is `/D`. No such path exists on Windows, so `decode_with_filesystem_check` returned `None` unconditionally.

There was no `cfg(windows)` and no `MAIN_SEPARATOR` anywhere in `utils.rs`.

## The fix

- the marker matches either separator
- the root split is the decoder's own job: `/` recovered from a leading dash, `D:\` from `D--`. Callers pass the folder name whole, which is what the two provider callers (`codebuddy`, `cursor_agent`) already did
- `decode_recursive_inner` takes a `&Path` base and joins, so no separator is ever written by hand

**The heuristic fallback stays Unix-only, deliberately.** Once the marker matches on Windows, `utils.rs:288-295` would start fabricating `/D--OneDrive/…` from a drive-lettered folder — strictly worse than today's behaviour of returning the storage path untouched. It is gated on the leading dash. This is also why the three defects had to land together rather than one at a time.

## The test helper was wrong too

```rust
let mut encoded = deep.to_string_lossy().replace(['/', '\\'], "-");
if !encoded.starts_with('-') { encoded.insert(0, '-'); }
```

It left the drive colon in and force-prepended a dash, producing `-C:-Temp-x` — a shape Claude Code never writes. So even with Windows in CI, these tests were asserting against a fiction. Now `['/', '\\', ':']`, no forced prefix, which reproduces both conventions exactly.

## What Windows users get back

Previously only `sessions-index.json` → `originalPath` worked. Without it:

- `decode_project_path` returned the storage path unchanged, so the raw `…\projects\D--OneDrive---…` folder surfaced as the project path
- `decode_project_path_verified` returned `None`, so #369's "the containing folder is authoritative, not a stale embedded `cwd`" never applied — a moved project kept showing its stale path

Callers are `commands/project.rs:313` (project list) and `commands/session/load.rs:614` (session loading).

Note the encoding is *more* ambiguous on Windows — spaces become `-` as well — so the filesystem-checked decoder is needed **more** there. Paths containing spaces remain ambiguous beyond what segment-splitting can resolve; that is upstream's anthropics/claude-code#7009 and out of scope here.

## Verification

Three tests. Two hold on any host — the marker is pure string matching, and the Unix root split is exercised locally. The drive-letter arm and the encode/decode round-trip are what the Windows job proves.

Mutation-checked rather than assumed:

| mutation | effect |
|---|---|
| marker back to `/`-only | marker test **fails** ✓ |
| Unix root split ignores the leading dash | 5 decode tests **fail** ✓ |

macOS: 998 and 1043 tests, both feature sets, green. Clippy `-D warnings` clean, fmt clean, tsc clean.

The one `#[allow]` added is `clippy::unnecessary_wraps` on `split_encoded_root` — clippy compiles a single arm and sees the Unix path always returning `Some`; the Windows arm genuinely returns `None`. Same shape as the existing `needless_return` carve-out in `commands::session::rename`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude session path handling across Windows and Unix-like systems.
  * Correctly processes Windows drive letters, slash types, and verbatim path prefixes.
  * Prevents Windows paths from being incorrectly interpreted as Unix-style paths.
  * Preserves accurate path decoding and round-trip behavior across supported platforms.
  * Improved handling for paths beginning with dashes.
  * Added more reliable support for recovering platform-specific project roots.

### DIFF
--- a/src-tauri/src/utils.rs
+++ b/src-tauri/src/utils.rs
@@ -98,15 +98,18 @@ pub fn find_line_starts(data: &[u8]) -> Vec<usize> {
 }
 
 pub fn extract_project_name(raw_project_name: &str) -> String {
-    if let Some(stripped) = raw_project_name.strip_prefix('-') {
-        // Prefer filesystem-checked decoding — handles arbitrarily deep paths
-        // and project names containing hyphens (e.g. ~/Projects/{Host}/{Org}/{Repo}).
-        if let Some(decoded) = decode_with_filesystem_check(stripped) {
-            if let Some(leaf) = Path::new(&decoded).file_name() {
-                return leaf.to_string_lossy().to_string();
-            }
+    // Prefer filesystem-checked decoding — handles arbitrarily deep paths
+    // and project names containing hyphens (e.g. ~/Projects/{Host}/{Org}/{Repo}).
+    // Passed whole: a Windows name is drive-lettered rather than dash-prefixed,
+    // so stripping first would have skipped it entirely (#548).
+    if let Some(decoded) = decode_with_filesystem_check(raw_project_name) {
+        if let Some(leaf) = Path::new(&decoded).file_name() {
+            return leaf.to_string_lossy().to_string();
         }
-        // Fallback: original heuristic for paths no longer on disk.
+    }
+    // Fallback: original heuristic for paths no longer on disk. Dash-prefixed
+    // (Unix-shaped) names only — see the note in `decode_project_path`.
+    if raw_project_name.starts_with('-') {
         let parts: Vec<&str> = raw_project_name.splitn(4, '-').collect();
         if parts.len() == 4 {
             return parts[3].to_string();
@@ -275,16 +278,17 @@ pub fn decode_project_path(session_storage_path: &str) -> String {
     }
 
     // 2. Fallback: decode from encoded directory name
-    const MARKER: &str = ".claude/projects/";
-    if let Some(marker_pos) = session_storage_path.find(MARKER) {
-        let encoded = &session_storage_path[marker_pos + MARKER.len()..];
-        if let Some(stripped) = encoded.strip_prefix('-') {
-            // Try filesystem-based decoding (recursive)
-            if let Some(path) = decode_with_filesystem_check(stripped) {
-                return path;
-            }
+    if let Some(encoded) = encoded_folder_name(session_storage_path) {
+        // Try filesystem-based decoding (recursive)
+        if let Some(path) = decode_with_filesystem_check(encoded) {
+            return path;
+        }
 
-            // Fallback: use heuristic decoding (still uses original encoded with dash)
+        // Fallback: the original heuristic. It builds a `/`-rooted path, so it
+        // only applies to a Unix-shaped name; on Windows it would fabricate
+        // `/D--OneDrive/...` from a drive-lettered folder, which is worse than
+        // returning the storage path untouched (#548).
+        if encoded.starts_with('-') {
             let parts: Vec<&str> = encoded.splitn(4, '-').collect();
             if parts.len() >= 4 {
                 return format!("/{}/{}/{}", parts[1], parts[2], parts[3]);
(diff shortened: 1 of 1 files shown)


jhlee0409:
## Windows: 33 → 30, no regressions, and the new tests pass there

| | baseline | now |
|---|---|---|
| run | 1007 | 1010 |
| passed | 974 | 980 |
| failed | 33 | **30** |

Fixed: `test_scan_projects_extracts_project_name`, `test_extract_project_name_deep_path`, `test_extract_project_name_with_hyphens_in_segments`. Nothing regressed.

More to the point, all three new tests pass on Windows — including `decode_project_path_resolves_a_real_folder_on_this_platform`, which decodes a real drive-lettered `C--Users-runneradmin-...` folder back to itself. That assertion could not have passed before this PR under any of the three defects.

## The round-trip test earned its place

It failed on the first Windows run, with exactly the failure the issue predicted:

```
left:  "//?/C--Users-runneradmin-AppData-Local-Temp-cchv_548_roundtrip-nested-leaf-dir"
right: "\\?\C:\Users\runneradmin\AppData\Local\Temp\cchv_548_roundtrip\nested\leaf-dir"
```

Two mistakes of mine, both invisible on macOS:

**Windows `canonicalize` returns verbatim paths** (`\\?\C:\...`). The fixture encoded that prefix, producing `--?-C--Users-...` — a shape Claude Code never writes, since it encodes the plain `C:\...` cwd. Stripped at the source now, plain and UNC, matching `strip_windows_extended_prefix` in rename.rs.

**The heuristic guard was defeated by it.** `encoded.starts_with('-')` assumed only a Unix name can lead with a dash; the verbatim prefix's leading `\` encodes to `-` too, so the `/`-rooted heuristic ran and fabricated that left-hand value.

My first correction — gating on `cfg(not(windows))` — overshot and broke three tests that were passing on Windows *through* the heuristic. Whether it applies is a property of the **name**, not the host: `-Users-jack-app` encodes a Unix path whichever OS reads it, and a drive-lettered `D--OneDrive-...` fails the dash test, which was the protection actually needed. Reverted to the shape test, with the fixture fixed at source so it holds.

Worth saying plainly: **the version first pushed here was wrong on Windows and green on macOS.** The only reason that surfaced before merge is a test written specifically because the platform cannot be reproduced locally.

## Remaining 30

Unrelated to this change — Unix path literals in fixtures and the known-folder sandbox gap, triaged on #541.

PR #550, opened 2026-08-28
fix(tests): give Windows fixtures platform-appropriate absolute paths

jhlee0409:
First pass at #541.

The dominant class: fixtures spelling absolute paths as Unix literals — `"/tmp/session.jsonl"`, `"/some/stale/Dev"`. **On Windows those are not absolute** (no drive), so the code under test rejected them with "path must be absolute" long before reaching the behaviour the test was about, and the assertion failed for the wrong reason.

Adds `test_utils::abs`, written Unix-style, producing `/tmp/x` or `C:\tmp\x` per host.

Two fixtures needed a directory that really exists and borrowed `/usr/lib`. Windows has none, so they build one now — `make_encoded_path` moves out of `utils.rs`'s test module into `test_utils`, since after #549 it encodes correctly for both platforms and two other modules want it.

`normalizes_windows_verbatim_path_prefix_for_root_comparison` asserted the mixed-case form while `normalize_path_for_comparison` **lowercases on Windows** — deliberately, documented on the function, because filesystem access there is case-insensitive and the allowlist boundary check has to match. A Windows-only test that had only ever run on Unix, where the `cfg` it covers is compiled out.

Also two separator-sensitive `contains("/sessions/")` checks in the codex tests, which cannot match a backslash path.

macOS 998 and 1043 green, clippy clean, fmt clean. **The Windows count is the point** — reporting it when the job finishes. Remaining classes (snapshots, the known-folder sandbox gap) are follow-ups on #541.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-platform reliability for path handling, including Windows compatibility.
  * Replaced system-specific paths with portable, platform-aware fixtures.
  * Corrected handling of Windows paths in session and provider data.
  * Added coverage for encoded project paths and platform-specific path casing.

* **Tests**
  * Centralized utilities for creating absolute, existing, and encoded paths.
  * Updated project scanning, session, provider, metadata, and path-traversal tests to use portable fixtures.
  * Improved snapshot coverage while preserving project path status validation.

### DIFF
--- a/src-tauri/src/commands/claude_settings.rs
+++ b/src-tauri/src/commands/claude_settings.rs
@@ -1070,7 +1070,8 @@ mod tests {
 
     #[test]
     fn test_validate_dialog_path_parent_dir_rejected() {
-        let path = Path::new("/some/path/../escape.txt");
+        let raw = crate::test_utils::abs("some/path/../escape.txt");
+        let path = Path::new(&raw);
         let result = validate_dialog_path(path);
         assert!(result.is_err());
         assert!(result.unwrap_err().contains("'..'"));
@@ -1078,7 +1079,8 @@ mod tests {
 
     #[test]
     fn test_validate_dialog_path_nonexistent_parent_rejected() {
-        let path = Path::new("/nonexistent_dir_abc123/file.txt");
+        let raw = crate::test_utils::abs("nonexistent_dir_abc123/file.txt");
+        let path = Path::new(&raw);
         let result = validate_dialog_path(path);
         assert!(result.is_err());
         assert!(result.unwrap_err().contains("does not exist"));
--- a/src-tauri/src/commands/metadata.rs
+++ b/src-tauri/src/commands/metadata.rs
@@ -377,7 +377,7 @@ mod tests {
 
     #[test]
     fn test_validate_project_metadata_key_absolute_path() {
-        assert!(validate_project_metadata_key("/tmp/project").is_ok());
+        assert!(validate_project_metadata_key(&crate::test_utils::abs("tmp/project")).is_ok());
     }
 
     #[test]
--- a/src-tauri/src/commands/project.rs
+++ b/src-tauri/src/commands/project.rs
@@ -537,7 +537,7 @@ mod tests {
                 "sessionId": "session-missing-worktree",
                 "timestamp": "2026-08-16T10:00:00Z",
                 "type": "user",
-                "cwd": "/tmp/deleted-worktree",
+                "cwd": crate::test_utils::abs("tmp/deleted-worktree"),
                 "message": { "role": "user", "content": "Keep this history" },
             })]),
         );
@@ -547,7 +547,10 @@ mod tests {
             .unwrap();
 
         assert_eq!(projects.len(), 1);
-        assert_eq!(projects[0].actual_path, "/tmp/deleted-worktree");
+        assert_eq!(
+            projects[0].actual_path,
+            crate::test_utils::abs("tmp/deleted-worktree")
+        );
 
         let serialized = serde_json::to_value(&projects[0]).unwrap();
         assert_eq!(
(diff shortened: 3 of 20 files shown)


jhlee0409:
## Windows: 41 → 5, no regressions

| pass | failing |
|---|---|
| baseline | 41 |
| absolute-path literals | 15 |
| JSON-embedded paths | 14 |
| separators + antigravity CLI sandbox | 7 |
| shared encoder + JSON escaping | 5 |

Nothing regressed at any step. macOS stayed at 998/1043 throughout.

### What the classes turned out to be

**Unix absolute literals (the bulk).** `"/tmp/x"` is not absolute on Windows, so the code rejected it with "path must be absolute" before reaching the behaviour under test — the assertion then failed for the wrong reason. `test_utils::abs` writes them Unix-style and produces the host's form.

**Paths inside JSON.** `abs` alone is not enough: `C:\tmp\x` has to be JSON-escaped, which a raw string cannot do. One of these was silently fatal — `test_load_project_sessions_prefers_jsonl_cwd_for_project_name` interpolated a real path into a raw JSONL line, making it unparseable, so the loader returned nothing and the test failed on the session count rather than the project name it was about.

**Expectations assembled with a literal `/`** while the value under test came from `Path::join`. Note `join("sessions/project-hash")` does not help — it keeps the slash verbatim; it has to be `join` per segment.

**A test encoder that was wrong.** `make_encoded_path` left the drive colon in and force-prepended a dash, producing `-C:-Temp-x`, a shape Claude Code never writes. Shared as `encode_path_claude_style` now, along with `strip_verbatim_prefix` — `canonicalize` returns `\\?\C:\...` on Windows and almost nothing else does, which defeated two separate comparisons.

**Two that were not fixtures at all.** `normalizes_windows_verbatim_path_prefix_for_root_comparison` asserted the mixed-case form while the function lowercases on Windows, deliberately and documented. And the project snapshots were missing `path_status` entirely, because `project_path_status` returns `None` for a non-absolute path — the snapshot diff was an absent field, not a different string.

**`antigravity_cli::default_root`** resolved `dirs::home_dir()` rather than the sandboxed helper, so all three CLI-layout tests scanned the real home on Windows. One line; it then exposed a stats test with no sandbox of its own, which is the #540 pattern.

### The 5 left, and why they are left

- `test_is_safe_path_downloads_accepted` / `_desktop_accepted` — `is_safe_path` builds its allowlist from `dirs::download_dir()`/`desktop_dir()`, which use the Windows known-folder API and consult neither `HOME` nor `CCHV_TEST_HOME`. Deliberate, for OneDrive redirection. Making these pass needs a test seam in a security function; asserting against the *real* Downloads instead would be the exact thing #536/#540 fixed.
- `test_global_summary_aggregates_reasoning_tokens`, `scan_projects_merges_desktop_and_cli_layouts` — both want `commands::antigravity::get_antigravity_root` sandboxed. I tried it: one line, and it cascades into **twelve** tests that reach the root with no sandbox. That is the module sweep described in #551, not a rider on a path-fixture PR.

All four belong to #551. Filed with counts: 22 provider modules, 29 direct `dirs::home_dir()` sites, 4 files using the sandboxed helper.

jhlee0409:
**Final: 41 → 4, no regressions.** The Gemini fix landed; `detect_session_provider_routes_new_providers` passes.

The remaining four are all blocked on #551, and both blockers are ones I verified rather than assumed:

- `test_is_safe_path_downloads_accepted` / `_desktop_accepted` — need a test seam in `is_safe_path`, which builds its allowlist from the Windows known-folder API by design. Asserting against the real Downloads instead would reintroduce exactly what #536/#540 fixed.
- `test_global_summary_aggregates_reasoning_tokens`, `scan_projects_merges_desktop_and_cli_layouts` — need `get_antigravity_root` sandboxed. I applied that one-line change and reverted it: it cascades into twelve tests with no sandbox of their own.

Ready to merge.

Issue #551, opened 2026-08-28
Provider layer bypasses the sandboxed home helper: 29 direct dirs::home_dir() sites, so provider tests are only isolated on Unix


jhlee0409:
Found while working #541. Not urgent, but it is the reason a whole class of provider tests only appear isolated.

## The gap

`crate::utils::home_dir()` was introduced in #537/#540 as the single home-resolution point: the real home normally, `CCHV_TEST_HOME` under `cfg(test)`, and a panic when a test reaches it without a sandbox. That last part is what made the audit possible.

The provider layer largely bypasses it:

| | count |
|---|---|
| provider modules calling `dirs::home_dir()` directly | **22** |
| call sites | **29** |
| provider files using the sandboxed helper | 4 |

`aider`, `claude`, `cline`, `codebuddy`, `codex`, `continue_dev`, `copilot_cli`, `crush`, `cursor`, `cursor_agent`, `forgecode` (×4), `gemini`, `goose`, `grok` (×2), `kimi` (×2), `kimi_code`, `ompi`, `opencode`, `openhands`, `openinterpreter`, `pi`, `qwen`, `vibe`, `vscode`.

## Why it has not bitten

On Unix, `dirs::home_dir()` reads `$HOME`, so a test that sets `HOME` does sandbox it — by coincidence, not by design. On Windows `dirs::home_dir()` goes through the known-folder API and consults neither `HOME` nor `USERPROFILE`, deliberately, to support OneDrive redirection.

So provider tests that set `HOME` are isolated on Unix and **silently resolve the developer's real home on Windows**.

That is exactly what #541 hit in `antigravity_cli`: all three CLI-layout tests set a fixture home, and on Windows the scan ran against the real one and returned nothing. Fixed there in #550, one line, and it immediately exposed a stats test reaching home with no sandbox at all — the #540 pattern again.

## Risk

Read-only for the most part; these are scan/load paths, so this is not the data-loss shape of #536/#540. The cost is that Windows provider coverage is not real, and the panic that makes unsandboxed access loud does not fire for any of these modules.

## Doing it

Mechanical per module, but 29 sites and each provider's tests carry their own home guard, so it should not ride along with unrelated work. `antigravity_cli` in #550 is a worked example including the fallout: converting one module surfaced one unsandboxed test that then needed a guard. Expect a similar tail per module.

Suggest one PR per few modules rather than a single sweep, so a cascade in one provider does not block the rest.

[...]

PR #555, opened 2026-08-29
fix(tests): sandbox the antigravity home, and share one guard

jhlee0409:
First module of #551. Picked first because two of #541's four remaining Windows failures are here.

## The gap

`commands::antigravity` resolved `dirs::home_dir()` at three sites. On Windows that goes to the known-folder API and consults neither `HOME` nor `CCHV_TEST_HOME`, so every test pointing `HOME` at a fixture was quietly scanning the developer's real home and asserting against an empty scan.

Converting it cascades into **thirteen** tests that reach the root with no sandbox of their own — which is why I reverted this out of #550 rather than riding it along.

## One guard instead of a seventh copy

There were already six guard definitions across the repo (`pi`, `openinterpreter`, `ompi`, `antigravity`, `stats`, `rename`), all slightly different. This adds `test_utils::SandboxHome` and migrates `providers::antigravity` onto it, deleting the local `HomeGuard`.

Two things the per-module guards did not do:

- **canonicalises** the path — on macOS the temp dir comes back as `/var/…` while anything resolving it sees `/private/var/…`. That is @nightcityblade's observation from #546, now applied at the source rather than patched per guard.
- **restores** the previous values rather than dropping them. Leaving `CCHV_TEST_HOME` set leaks the sandbox forward into every later test in the same process — which is precisely what hid nine unsandboxed tests until CI ran them under nextest with a process each (#544).

Five definitions remain; they migrate as their modules are converted.

## Verification

| path | result |
|---|---|
| `cargo test -- --test-threads=1` (release checklist) | 1043 / 0 failed |
| `cargo nextest run --profile ci` | 998 passed |
| `… --features webui-server` | 1043 passed |

## One thing not fixed, stated plainly

Plain `cargo test` with **default parallelism** is not green — and I made it worse.

```
develop:    1010 passed, 15 failed
this branch: 1002 passed, 23 failed
```

`#[serial]` only orders serial tests against each other; untagged tests still run alongside them, so adding thirteen more HOME mutators widens the window. develop is already broken there for the same reason.

That configuration is neither the CI path (nextest, process per test) nor the release path (`--test-threads=1`), and it only closes properly when every HOME-toucher is tagged — which is the end state #551 describes. Flagging it rather than leaving it to be discovered.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved home-directory detection so configured `HOME` locations are honored consistently across platforms.
  * Prevented usage and statistics data from being read from unintended local directories.
  * Improved reliability when loading usage records and conversation statistics from configured locations.

* **Tests**
  * Improved test isolation by using temporary home directories.
  * Reduced test interference when running home-directory and usage-related tests in parallel.

### DIFF
--- a/src-tauri/src/commands/antigravity.rs
+++ b/src-tauri/src/commands/antigravity.rs
@@ -35,7 +35,10 @@ static MODEL_ALIAS_MAP: Lazy<HashMap<&'static str, &'static str>> = Lazy::new(||
 
 /// 定位 antigravity 根目录：`~/.gemini/antigravity`
 pub fn get_antigravity_root() -> Option<PathBuf> {
-    dirs::home_dir().map(|h| h.join(".gemini").join("antigravity"))
+    // Sandboxed helper, not `dirs::home_dir()`: on Windows the latter goes to
+    // the known-folder API and ignores `HOME`, so a test pointing `HOME` at a
+    // fixture was silently scanning the developer's real one (#551).
+    crate::utils::home_dir().map(|h| h.join(".gemini").join("antigravity"))
 }
 
 /// Resolves the antigravity root directory, with fallback discovery logic.
@@ -110,7 +113,7 @@ pub fn get_antigravity_rpc_cache_root(root: &Path) -> PathBuf {
 fn discover_external_state_dirs() -> Vec<PathBuf> {
     let mut dirs = Vec::new();
     let bases = if cfg!(target_os = "macos") {
-        dirs::home_dir()
+        crate::utils::home_dir()
             .map(|home| vec![home.join("Library").join("Application Support")])
             .unwrap_or_default()
     } else if cfg!(target_os = "windows") {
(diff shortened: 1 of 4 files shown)


jhlee0409:
## Windows 4 → 2, and the last two are one problem

Both antigravity failures are gone. What remains is `test_is_safe_path_downloads_accepted` / `_desktop_accepted` — the known-folder seam, which is a different fix in a security function.

## Two things Windows CI caught that macOS could not

**The canonicalisation was half a fix.** Canonicalising resolves the macOS `/var` → `/private/var` divergence, but on Windows it returns the extended-length form, so the sandbox was exporting `CCHV_TEST_HOME=\\?\C:\Users\…`. The desktop root survived that — it is built by joining onto whatever home says — but the CLI root vanished, because code comparing plain paths cannot match a verbatim one. `strip_verbatim_prefix` already existed for exactly this from #550; applied at the source now.

**Then the same #541 shape one level up.** The merge test asserted on the literal `antigravity-cli:///tmp/cli-proj` while its fixture had already moved to a host-appropriate path, so it went looking for a Unix workspace. Both sides now derive from one `abs` call.

Neither was visible locally: macOS canonicalisation is plain, and the Unix literal is correct on Unix.

## Final

| path | result |
|---|---|
| `cargo test -- --test-threads=1` | 1043 / 0 |
| Windows (informational) | **2 failed**, down from 4 |

The default-parallelism caveat in the description still stands and is unchanged by these two commits.

PR #556, opened 2026-08-29
fix(tests): route the provider layer through the sandboxed home

jhlee0409:
Second and largest part of #551, after #555 did antigravity.

**28 `dirs::home_dir()` call sites across 23 provider modules** now go through `crate::utils::home_dir()`.

## Why it matters only on Windows

On Unix the two are equivalent — `dirs::home_dir()` reads `$HOME` and the tests set it, so the sandbox worked by coincidence. On Windows `dirs::home_dir()` goes through the known-folder API and consults neither `HOME` nor `CCHV_TEST_HOME`, deliberately, to support OneDrive redirection.

So every provider test pointing `HOME` at a fixture was resolving the **developer's real home** there, and the `cfg(test)` panic that makes an unsandboxed read loud never fired for any of these modules.

## The tail

Fifteen tests were reaching home with no sandbox of their own — far fewer than the 28 sites suggested. All now hold `SandboxHome`.

One was worse than merely unguarded. `test_safe_session_path_allows_kimi_sessions` saved and restored `HOME` around its own `TempDir`:

```rust
let temp = TempDir::new().unwrap();
std::env::set_var("HOME", temp.path());
// … fixture written under temp …
```

which leaves `CCHV_TEST_HOME` pointing somewhere else entirely — the path the code under test actually resolves. It passed because on Unix `HOME` was enough. The guard owns the home now and the fixture lives inside it.

## Not included

Five guard definitions remain (`pi`, `openinterpreter`, `ompi`, `stats`, `rename`). Migrating them onto `SandboxHome` means retargeting each fixture at the guard's own directory rather than a separately-created temp — a per-call-site rewrite with real room to get wrong quietly. Following separately.

## Verification

| path | result |
|---|---|

Windows count reported when the job finishes — that is what this is for.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of session and project data across supported AI tools, including customized home-directory locations.
  * Increased reliability when locating provider-specific folders and validating session paths.
  * Reduced environment-dependent behavior that could cause incorrect results.

* **Tests**
  * Strengthened isolation and coverage for session deletion, path validation, provider detection, and project discovery.
  * Improved consistency across platforms and home-directory configurations.

### DIFF
--- a/src-tauri/src/commands/session/delete.rs
+++ b/src-tauri/src/commands/session/delete.rs
@@ -97,6 +97,7 @@ pub async fn delete_session(file_path: String) -> Result<(), String> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use serial_test::serial;
     #[cfg(unix)]
     use std::os::unix::fs::symlink;
     use tempfile::TempDir;
@@ -162,7 +163,9 @@ mod tests {
     }
 
     #[tokio::test]
+    #[serial]
     async fn trash_valid_jsonl_file() {
+        let _home = crate::test_utils::SandboxHome::new();
         let dir = TempDir::new().unwrap();
         let file = dir.path().join("abc-123.jsonl");
         fs::write(&file, "{}\n").unwrap();
@@ -172,7 +175,9 @@ mod tests {
     }
 
     #[tokio::test]
+    #[serial]
     async fn trash_jsonl_and_associated_directory() {
+        let _home = crate::test_utils::SandboxHome::new();
         let dir = TempDir::new().unwrap();
         let file = dir.path().join("session-1.jsonl");
         let assoc_dir = dir.path().join("session-1");
--- a/src-tauri/src/commands/session/mod.rs
+++ b/src-tauri/src/commands/session/mod.rs
@@ -168,6 +168,7 @@ mod tests {
     #[test]
     #[serial]
     fn accepts_session_under_symlinked_claude_root() {
+        let _home = crate::test_utils::SandboxHome::new();
         let tmp = tempfile::tempdir().expect("tempdir");
         let home = tmp.path();
         let store_projects = home.join(".claude-store").join("projects").join("proj");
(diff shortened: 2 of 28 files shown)


jhlee0409:
## Done: 28 sites, 23 modules, and the guards collapsed to one

Windows stays at **2 failures**, both the known-folder seam — nothing else in the provider layer is left resolving the real home.

### What CI caught that macOS could not

**Ubuntu found a test that was vacuous on my machine.** `grok::get_base_path_returns_none_when_default_dir_absent` bails out when `~/.grok` exists:

```rust
if dirs::home_dir().map(|h| h.join(".grok").exists()).unwrap_or(false) { return; }
```

I have one, so it returned before asserting and reported ok. The runner does not, so it ran on and panicked for want of a sandbox. Both halves fixed: it holds `SandboxHome`, and the skip is **deleted** — it only existed because the real home might contain the directory under test, and inside a sandbox it cannot. `kimi` carried the identical shape and loses its skip too.

Mutation-checked: making `get_base_path` return `Some` now fails it. Before, it passed on any machine with a `~/.grok` whatever the function did.

**Windows found a fixture built from the wrong home.** `extract_workspace_display_name_prefers_cwd_basename_and_ignores_home_dir` constructed its "this cwd is the home, ignore it" entry with `dirs::home_dir()` — the real home there, while the code resolves the sandbox. So the entry it was meant to ignore was not the home at all.

### Guards

Four near-copies (`pi`, `ompi`, `stats`, `rename`) are gone; 56 call sites hold `SandboxHome`. `openinterpreter` keeps its `HomeGuard` — despite the name it swaps `INTERPRETER_HOME`, a provider override, not the home.

### Verification

| path | result |
|---|---|
| Windows (informational) | 2 failed, unchanged |

Also worth noting: three of the four bugs in this PR were only visible on a platform I cannot run. The pattern is consistent enough now to be worth stating — a test that consults the real environment to decide whether to assert is a test that reports ok for the wrong reason somewhere.

PR #557, opened 2026-08-29
fix(tests): sandbox the commands layer, and give is_safe_path a test seam

jhlee0409:
Finishes the #551 sweep started in #555 (antigravity) and #556 (providers). Twelve more sites across `claude_settings`, `mcp_presets`, `unified_presets`, `project`, `settings` and `session`. **The provider layer is already at zero.**

## The one that matters: `is_safe_path`

This is the last of #541's Windows failures, and it needed a decision rather than a guard.

`is_safe_path` builds its allowlist from `dirs::download_dir()`, `document_dir()` and `desktop_dir()`. Those go through the Windows known-folder API **deliberately**, so a Downloads redirected into OneDrive is still honoured — and they consult neither `HOME` nor `CCHV_TEST_HOME`.

That left the function untestable there. The allowlist named the developer's real Downloads, so the only way to exercise the *accept* path would have been to write into it — which is precisely the mistake #536 and #540 were about.

```rust
fn known_folder(lookup: fn() -> Option<PathBuf>) -> Option<PathBuf> {
    if cfg!(test) { None } else { lookup() }
}
```

Under test each entry collapses to its home-relative fallback, so the whole allowlist derives from the sandboxed home. Production is untouched.

**What this gives up:** coverage of the redirection branch itself. That branch is a single OS call with nothing to assert about — the behaviour worth testing is whether a path inside or outside the allowlist is accepted, and that is now testable on every platform instead of none.

## Fixtures that were pointing at the wrong home

Three tests created their own temp and pointed `HOME` at it, leaving `CCHV_TEST_HOME` — the path actually resolved — somewhere else:

- `accepts_session_under_symlinked_claude_root`, via a `with_home` helper
- the two `server::tests` restore cases, which had no sandbox at all

## Verification

| path | result |
|---|---|

Mutation-checked the seam, since the whole claim is that the accept path is now real: emptying `allowed_dirs` after the loop **fails both accept tests**. Before this they could not run on Windows at all.

If Windows comes back at zero, #541 closes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when locating home directories and application folders across environments.
  * Enhanced safe-path validation for settings, projects, sessions, and presets, including symlinked locations.

* **Tests**
  * Improved test isolation and consistency for home-directory and restore-related scenarios.

### DIFF
--- a/src-tauri/src/commands/claude_settings.rs
+++ b/src-tauri/src/commands/claude_settings.rs
@@ -41,12 +41,12 @@ pub struct AllMCPServers {
 
 /// The home directory these settings paths hang off.
 ///
-/// In a normal build this is exactly `dirs::home_dir()`.
+/// In a normal build this is exactly `crate::utils::home_dir()`.
 ///
 /// # Why this indirection exists
 ///
 /// The tests below intended to sandbox themselves with
-/// `env::set_var("HOME", temp_dir)`. That is inert on Windows: `dirs::home_dir()`
+/// `env::set_var("HOME", temp_dir)`. That is inert on Windows: `crate::utils::home_dir()`
 /// resolves through the known-folder API and consults neither `HOME` nor
 /// `USERPROFILE`. So `test_save_and_retrieve_user_settings` wrote
 /// `{"theme":"dark","fontSize":14}` straight over the developer's real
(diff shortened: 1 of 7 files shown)


jhlee0409:
## Windows: **1010 / 1010, zero failures**

```
Summary [ 11.044s] 1010 tests run: 1010 passed, 5 skipped
```

Down from 41 when the platform first entered CI in #539. The seam did what it was for — both `is_safe_path` accept tests now run there, and mutation-checking confirms they discriminate (emptying `allowed_dirs` after the loop fails both).

That makes the whole Rust suite green on Ubuntu, macOS and Windows for the first time.

PR #568, opened 2026-09-03
fix(sessions): preserve branched Claude conversations

chaosink:
## What & why

Claude Code can create multiple continuation session files that reference the same logicalParentUuid. The original session may also continue receiving messages after that branch point.

The existing cross-file continuation logic treated every predecessor as superseded, which could incorrectly hide the original conversation from the session list.

This change:

- keeps a predecessor visible when multiple direct continuations reference it
- keeps it visible when main-conversation messages continue after the referenced UUID
- preserves the existing behavior for unambiguous linear continuation chains
- only affects Claude Code cross-file session handling

## Type

- [x] Bug fix
- [ ] New feature
- [ ] New provider (Claude Code / Codex / OpenCode / Kiro / Kimi / Copilot / ...)
- [ ] Refactor / perf
- [ ] Docs

## Checklist

- [x] PR targets **develop**, not main
- [x] Added/updated **tests** for the changed behavior
- [x] pnpm exec tsc --build . and pnpm lint pass
- [x] **i18n**: N/A - no new user-facing strings

## Testing

- cargo test --manifest-path src-tauri/Cargo.toml commands::session::chain::tests --lib
- cargo fmt --manifest-path src-tauri/Cargo.toml -- --check
- cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets --all-features -- -D warnings
- pnpm exec tsc --build .
- pnpm lint

All 10 session-chain tests pass. Lint completes with 0 errors; its single warning is in an unchanged file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversation chain handling when branches are created.
  * Parent conversations now remain visible when they have multiple branches or continued main-conversation activity.
  * Linear predecessor conversations are hidden more accurately when fully superseded by a single successor.
  * Sidechain messages are correctly distinguished from main-conversation messages when determining conversation visibility.

PR #572, opened 2026-09-04
fix(sessions): hide redundant Claude predecessors

chaosink:
## What & why

Thank you for reviewing and merging #568 so quickly.

After the merge, I noticed an edge case I had overlooked: multiple direct continuations alone should not keep an otherwise inactive predecessor visible. Sorry for missing this in the original change.

When a predecessor has no messages after the referenced branch points, each child session already includes the shared predecessor history. Showing the predecessor separately therefore creates a redundant session-list entry.

This follow-up:

- hides a predecessor when it has no main-conversation messages after any referenced branch point, regardless of the number of direct children
- continues to show the predecessor when it has user or assistant messages after at least one referenced branch point
- deduplicates repeated branch-point UUIDs before inspecting the predecessor
- remains scoped to Claude Code cross-file continuation handling

## Type

- [x] Bug fix
- [ ] New feature
- [ ] New provider (Claude Code / Codex / OpenCode / Kiro / Kimi / Copilot / ...)
- [ ] Refactor / perf
- [ ] Docs

## Checklist

- [x] PR targets **develop**, not main
- [x] pnpm exec tsc --build . and pnpm lint pass
- [x] **i18n**: N/A - no new user-facing strings

## Testing

- cargo test --manifest-path src-tauri/Cargo.toml commands::session::chain::tests --lib - 11 passed
- cargo test --manifest-path src-tauri/Cargo.toml commands::session:: --lib -- --test-threads=1 - 186 passed
- pnpm exec vitest run - 1200 passed
- pnpm exec tsc --build .
- pnpm lint
- cargo fmt --manifest-path src-tauri/Cargo.toml -- --check
- cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets --all-features -- -D warnings

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session history handling for branched conversations.
  * Inactive parent sessions are now hidden when none of their branches contain subsequent main-conversation activity.
  * Active parent sessions with multiple branches remain visible.

### DIFF
--- a/src-tauri/src/commands/session/chain.rs
+++ b/src-tauri/src/commands/session/chain.rs
@@ -251,13 +251,13 @@ pub fn resolve_session_chain(session_path: &Path) -> Vec<PathBuf> {
     chain
 }
 
-/// Every file in `project_root` that is unambiguously replaced by one direct
-/// continuation. A predecessor remains visible when multiple files branch
-/// from the same boundary, or when it keeps receiving conversation turns
-/// after the referenced UUID. Those shapes are forks, not linear automatic
-/// continuation chains.
+/// Every file in `project_root` whose direct continuations already cover its
+/// conversation history. Multiple children can share the same predecessor
+/// without making that predecessor useful as a separate session-list entry.
+/// It remains visible only when it kept receiving main-conversation turns
+/// after at least one referenced branch point.
 pub fn superseded_chain_paths(project_root: &Path) -> HashSet<PathBuf> {
-    let mut direct_successors: HashMap<PathBuf, Vec<String>> = HashMap::new();
+    let mut referenced_branch_points: HashMap<PathBuf, HashSet<String>> = HashMap::new();
     let Ok(entries) = fs::read_dir(project_root) else {
         return HashSet::new();
     };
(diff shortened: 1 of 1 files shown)


jhlee0409:
Thanks for the fast follow-up, and for being upfront about the edge case in #568.

I traced this against #568 and confirmed:
- The branch is current (built on develop's HEAD, after both #568 and the symlink-skip fix).
- Evaluating every distinct referenced branch-point UUID with `.any(...)` is more precise than #568's `parent_uuids.len() == 1` proxy: a parent with its own activity after any branch point stays visible, and two children branching from the exact same UUID are no longer treated as "always keep".
- Opening any child still merges the parent's messages via `resolve_session_chain`, so hiding the parent from the list doesn't lose content.
- I ran the chain test module locally: all 12 tests pass.

One thing I want to settle deliberately before merging: this removes one of #568's two independent "keep visible" rules (multiple direct continuations → always visible). That's a user-visible change to the session list rather than a tightened edge case — `superseded_paths_keeps_a_parent_with_multiple_direct_children` is inverted in this PR — so I'd like to confirm that policy on our side before merging. I'll follow up here.

Non-blocking: a test with two *different* branch-point UUIDs (one with trailing activity, one without) would pin the new `.any()` behavior more precisely than the current single-UUID cases.

PR #573, opened 2026-09-07
feat(provider): add Z Code (Z.ai) session source

zhibinlei:
## What & why

Adds [Z Code](https://zcode.z.ai) — Z.ai's GLM-powered coding agent — as a new session source, so its conversations show up alongside Claude Code / Codex / etc. No open issue; this closes the gap for Z Code users directly.

Z Code stores its authoritative transcript history in a single SQLite DB at `~/.zcode/cli/db/db.sqlite` (WAL) with three tables: `session` (cwd, generated title, task_type), `message` (role, model, token usage) and `part` (ordered content blocks). A `ZCODE_HOME` env override follows the existing `CODEX_HOME`/`GEMINI_HOME` convention.

### Implementation notes

- **New `providers/zcode.rs`** following the existing provider pattern (rusqlite read-only + 5s busy_timeout, `zcode://<dir>#<session_id>` pseudo-paths like the Crush provider). Projects group by the session `directory`; session summaries use the DB's own generated titles.
- **Tool call/result split**: Z Code merges a call and its result into a single `tool` part (`state.input` / `state.output`). The provider emits a `tool_use` block on the assistant message plus a synthesized user-lane `tool_result` message, mirroring Claude Code transcripts so the existing tool-card rendering works unchanged.
- `reasoning` parts → `thinking` blocks; `step-start`/`step-finish` markers, `timeline` events and `zcode-artifact://` file attachments are skipped. Subagent sessions (`task_type = 'subagent_child'`) are excluded, matching sidechain handling elsewhere.
- Token usage maps with the cached subset kept out of `input_tokens` (same convention as the Qwen provider) so totals don't double-count cache reads.
- Global search narrows candidate sessions in SQL (`EXISTS … part.data LIKE ? ESCAPE '\'`, ASCII-safe queries only, metacharacters escaped) before any JSON parsing; non-ASCII/unsafe queries keep the full-scan path where the Unicode-aware Rust matcher stays the only judge.
- Wired into scan / sessions / messages / search dispatch, the three default provider fallback lists, stats routing (project + `zcode://` session scheme detection, with the `#<session_id>` suffix stripped from display names), the WebUI session-path allowlist, the file watcher, the frontend provider registry (types, provider tabs, i18n label, colour) and all five READMEs (provider count 29 → 30).
- Reuses the `common.provider.zcode` i18n key that #450 pre-seeded (value normalised to the product's own styling "Z Code").

## Type

- [ ] Bug fix
- [ ] New feature
- [x] New provider (Claude Code / Codex / OpenCode / Kiro / Kimi / Copilot / …)
- [ ] Refactor / perf
- [ ] Docs

## Checklist

- [x] **i18n**: any new user-facing string is `t()`-wrapped and the key exists in **all 5 languages** (`en, ko, ja, zh-CN, zh-TW`) — verified with `pnpm run i18n:validate`

No new commands — the provider plugs into the existing multi-provider scan/sessions/messages/search dispatch.

- [x] Followed the existing provider pattern in `src-tauri/src/providers/`
- [x] Session discovery verified on **Linux (WSL2)** against a live `~/.zcode` store (4 projects / 5 sessions, end-to-end through the WebUI API: detect → scan → sessions → messages → search) and on **Windows 11** where the same layout was confirmed at `C:\Users\<user>\.zcode\cli\db\db.sqlite`. Not verified on macOS (no device available) — the store path is the same `~/.zcode` convention on all three platforms.
- [x] No leftover identifiers copied from another provider (names, comments, fixtures)

### Testing

- Provider unit tests (11): scan grouping + subagent/NULL-`task_type` exclusion handling, session metadata, tool call → `tool_use`/`tool_result` split (blocks, usage arithmetic, duration, parent linkage), session-path parsing, LIKE-pattern escaping contract, ASCII prefilter match, non-ASCII full-scan fallback, literal-`%` behaviour, and the public API (`detect`/`scan_projects`/`load_sessions`/`load_messages`/`search`) against a real file-backed `db.sqlite` in a TempDir via `ZCODE_HOME`, plus a missing-store case.
- Full serial Rust suite: **1017/1017**; `clippy --all-targets --all-features -D warnings` and `fmt --check` clean; `tsc`, `vitest` **1200/1200**, `lint` (0 errors) and `i18n:validate` pass.

## Summary by CodeRabbit

- **New Features**
  - Added support for detecting and browsing Z Code projects, sessions, messages, and conversation search.
  - Added Z Code to provider filters, statistics, project navigation, and file monitoring.
  - Added localized “Z Code” labels across supported languages.
  - Added support for configuring the Z Code home directory.

- **Bug Fixes**
  - Improved handling of incomplete Z Code session data and session paths.
  - Improved provider-specific error reporting during Z Code searches and project loading.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -70,7 +70,7 @@ Docker、VPS、systemdのセットアップは[サーバーモード](#サーバ
 
 AIコーディングアシスタントは数千もの会話メッセージを生成しますが、ツール間で履歴を振り返る方法を提供していません。CCHVがこの課題を解決します。
 
-**29のアシスタント。1つのビューア。** Claude Code、GitHub Copilot、Gemini CLI、Antigravity、Codex CLI、Cline（Roo Code・Kilo Code含む）、Cursor、Cursor Agent、Aider、OpenCode、ForgeCode、CodeBuddy Code、Grok CLI、Kimi、Kiro、Amazon Q CLI、Continue.dev、PearAI、Goose、Crush、llm、Open Interpreter、Pi、oh-my-pi、Mistral Vibe、Qwen Code、Zed、OpenHands、Traeのセッションをシームレスに切り替え — トークン使用量を比較し、プロバイダー間で検索し、ワークフローを1つのインターフェースで分析。
+**30のアシスタント。1つのビューア。** Claude Code、GitHub Copilot、Gemini CLI、Antigravity、Codex CLI、Cline（Roo Code・Kilo Code含む）、Cursor、Cursor Agent、Aider、OpenCode、ForgeCode、CodeBuddy Code、Grok CLI、Kimi、Kiro、Amazon Q CLI、Continue.dev、PearAI、Goose、Crush、llm、Open Interpreter、Pi、oh-my-pi、Mistral Vibe、Qwen Code、Zed、OpenHands、Trae、Z Codeのセッションをシームレスに切り替え — トークン使用量を比較し、プロバイダー間で検索し、ワークフローを1つのインターフェースで分析。
 
 | プロバイダー | データの場所 | 取得できる情報 |
 |----------|--------------|--------------|
@@ -101,6 +101,7 @@ AIコーディングアシスタントは数千もの会話メッセージを生
 | **Mistral Vibe** | `~/.vibe/logs/session/` | reasoningとツール呼び出しを含むOpenAIスタイルのチャットトランスクリプト（`VIBE_HOME`で上書き可） |
 | **Qwen Code** | `~/.qwen/projects/.../chats/` | セッション単位のJSONLトランスクリプト（ツール呼び出し、思考プロセス、トークン使用量） |
 | **Zed** | `…/Zed/threads/threads.db` | Agent Panelスレッド — SQLite + Zstd圧縮JSON |
+| **Z Code** | `~/.zcode/cli/db/db.sqlite` | Z.ai の GLM コーディングエージェント — session/message/part 型 SQLite ストア（タイトル、ツール呼び出し、思考、トークン使用量） |
 | **OpenHands** | `~/.openhands/sessions/` | クラシックなイベントストア形式の会話 |
 | **Trae** | `…/Trae/User/workspaceStorage/.../state.vscdb` | ワークスペース単位のチャット（icubeストア；実験的、リバースエンジニアリングによる対応） |
 
@@ -128,7 +129,7 @@ Antigravityに関する注記：ビューアはAntigravityルートを`~/.gemini
 
 | 機能 | 説明 |
 |---------|-------------|
-| **マルチプロバイダー対応** | **29のAIコーディングアシスタント**を統合ビューアで閲覧 — Claude Code、GitHub Copilot、Gemini CLI、Codex CLI、Cursor / Cursor Agent、Cline（Roo Code・Kilo Code含む）、Aider、OpenCode、ForgeCode、CodeBuddy Code、Grok CLI、Kimi、Kiro、Antigravity、Amazon Q CLI、Continue.dev、PearAI、Goose、Crush、llm、Open Interpreter、Pi、oh-my-pi、Mistral Vibe、Qwen Code、Zed、OpenHands、Trae — プロバイダー別フィルタリング、ツール間比較 |
+| **マルチプロバイダー対応** | **30のAIコーディングアシスタント**を統合ビューアで閲覧 — Claude Code、GitHub Copilot、Gemini CLI、Codex CLI、Cursor / Cursor Agent、Cline（Roo Code・Kilo Code含む）、Aider、OpenCode、ForgeCode、CodeBuddy Code、Grok CLI、Kimi、Kiro、Antigravity、Amazon Q CLI、Continue.dev、PearAI、Goose、Crush、llm、Open Interpreter、Pi、oh-my-pi、Mistral Vibe、Qwen Code、Zed、OpenHands、Trae、Z Code — プロバイダー別フィルタリング、ツール間比較 |
 | **会話ブラウザ** | プロジェクト/セッション別に会話を閲覧（ワークツリーグループ化対応） |
 | **グローバル検索** | 全プロバイダーの会話を瞬時に検索 |
 | **分析ダッシュボード** | デュアルモードトークン統計（課金 vs 会話）、コスト内訳、プロバイダー分布チャート |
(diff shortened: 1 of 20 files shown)


zhibinlei:
Thanks for the review — all three findings addressed in 242c77e:

1. **Default provider lists** (`stats.rs`, `multi_provider.rs`): `zcode` added to `all_stats_providers()`, the `scan_all_projects` fallback, and the `search_all_providers` fallback. Verified end-to-end: a `scan_all_projects` call with no `activeProviders` now returns Z Code projects alongside Claude's.
2. **`zcode://` stats routing**: the `path_under_root` branch is replaced by a scheme check in `detect_session_provider`'s URI chain (matching `crush://` / `qwen://` etc.); also covered by a new assertion in `detect_session_provider_routes_new_providers`.
3. **`session_count` consistency**: `scan_projects_conn` now filters `WHERE s.m_cnt > 0` so the project row matches what `load_sessions` shows; the scan test asserts the message-less session is excluded, and the provider-count guard test moves 29 → 30.

Also: commit history was rewritten to use a GitHub noreply author address.

zhibinlei:
Round 2 addressed in 70dc8d0:

1. **Session project name suffix** (`stats.rs`): `resolve_provider_project_name_from_session` now strips the `#<session_id>` suffix for Z Code pseudo-paths before taking the last path segment — `zcode:///home/jack/proj#sess-1` reports `proj`, covered by a new unit test. (Noted while fixing: Crush session pseudo-paths share the same `<dir>#<id>` shape, so the generic fallback likely has the same cosmetic issue there — happy to extend the strip in a follow-up if you prefer it fixed for both.)
2. **Search prefilter** (`zcode.rs`): candidate sessions are now narrowed in SQL before any JSON parsing — an indexed `EXISTS (... part.data LIKE ? ESCAPE '\')` probe with `%`/`_`/`\` escaped. The prefilter only applies to queries that are pure ASCII without quotes, backslashes or control characters (SQLite `LIKE` is ASCII-case-insensitive and raw `part.data` is JSON, so anything else could lose matches); non-ASCII and otherwise unsafe queries keep the full-scan path where the Unicode-aware Rust matcher stays the only judge. The probe rides the existing `part_session_idx` — no new index needed (and the store is opened read-only, so none could be added).
3. Tests added: LIKE-pattern escaping contract, ASCII prefilter match, non-ASCII full-scan fallback, literal-`%` behaviour, and the session-name strip. Serial `cargo test` (zcode + stats): 89/89; `clippy --all-features -D warnings` clean; verified end-to-end on a live store (ASCII and CJK queries both return hits).

zhibinlei:
Follow-up review round addressed — the branch was also rebased onto `develop` and the PR retargeted accordingly:

1. **Base branch**: now targets `develop` (rebased cleanly over the 35 commits; no conflicts).
2. **`ZCODE_HOME` override**: store root honours `ZCODE_HOME` (relative values resolve against the cwd), falling back to `~/.zcode` — same convention as `CODEX_HOME`/`GEMINI_HOME`, following the vibe provider's implementation.
3. **Public API coverage**: `detect`, `get_base_path`, `scan_projects`, `load_sessions`, `load_messages` and `search` now run against a real file-backed `db.sqlite` in a TempDir via `ZCODE_HOME`, with the vibe-style env save/restore; plus a missing-store case.
4. **Null safety**: `title`/`title_source` read as `Option<String>` (no more silent row drops via `.flatten()`); all three queries guard `task_type` with `(IS NULL OR != 'subagent_child')`. The observed ZCode 3.11 schema declares these columns NOT NULL — documented in the module docs, with the queries degrading gracefully anyway. A NULL-`task_type` test row proves inclusion.
5. **Platform evidence**: store layout confirmed identical on Linux (WSL2, live end-to-end) and Windows 11 (`C:\Users\<user>\.zcode\cli\db\db.sqlite`); stated in the PR description. macOS unverifiable here (no device) — path is the same `~/.zcode` convention.
6. Cleanups: unreachable backslash arm removed from the LIKE escaper, scan rows collected directly (SQL already emits one per directory), `zcode` sorted before `zed` in the frontend registries, doc comments added to the private helpers, and the vestigial allowlist entry kept with a symmetry comment.

Verification: full serial Rust suite **1017/1017**, `clippy --all-targets --all-features -D warnings` clean, `fmt --check` clean, `tsc` clean, `vitest` **1200/1200**, `lint` 0 errors, `i18n:validate` pass. Commit trailers corrected to `Co-Authored-By: Claude Code`.

jhlee0409:
Thanks for the thorough work here, and for chasing all three CodeRabbit rounds down to root cause rather than just silencing the linter — the default-provider-list fix, the `zcode://` scheme routing, and the search-prefilter safety analysis (ASCII-only LIKE probe, full Unicode-aware fallback for everything else) all check out on re-read.

**CI:** `Frontend Tests` and `Rust Tests` were sitting at `action_required` on this fork PR, so the reported test/clippy/fmt results hadn't been verified by CI yet. I've approved those runs on the current head — I'll merge once they're green.

Minor: the round-3 note says the unreachable backslash arm was dropped from `like_prefilter_pattern`'s escaper, but it's still in the diff (`'\\' => "\\\\".to_string()`). Harmless dead code since backslash queries are excluded upstream — just flagging the mismatch.

Non-blocking notes for a follow-up if you're interested:
- `has_errors` is hardcoded `false` in `load_sessions_conn` even though `is_error` is computed per tool part a few lines up, so the session list won't show a failure badge for sessions whose only failure was a tool error.
- `detect()` returns `None` (rather than `Some(..., is_available: false)`) when `~/.zcode` exists without `db.sqlite` yet — slightly different from `codex.rs`'s "installed but empty" handling. Low-stakes UX drift, not a correctness issue.
- Z Code's store uses OpenCode's schema vocabulary (`modelID`/`parentID`/`callID`/`reasoning`/`step-finish`), so the tool-call split / reasoning mapping / usage math could eventually share a helper with `opencode.rs`. Consistent with existing precedent, so not something I'd block on.

jhlee0409:
CI update: `Rust Tests` passes on ubuntu (tests + clippy/fmt), but the **Windows** job fails one test:

```
providers::zcode::tests::public_api_round_trip_via_zcode_home
panicked at src\providers\zcode.rs:859:9:
assertion failed: get_base_path().expect("base path").ends_with("cli/db")
```

On Windows `get_base_path()` returns `...\cli\db`, so the string `ends_with("cli/db")` check fails. Comparing path components instead works on every platform, e.g.:

```rust
let base = get_base_path().expect("base path");
assert!(std::path::Path::new(&base).ends_with(std::path::Path::new("cli").join("db")));
```

(The `Frontend Tests` failure here was a pre-existing develop issue — a jsdom/nwsapi regression — now fixed by #584; please merge or rebase on `develop` when you push the fix so the frontend job picks it up.) Once Windows is green I'll merge.

PR #578, opened 2026-09-10
feat(stats): roll up spawned subagent sessions into session token stats

### LINKED ISSUE #577: feat(stats): roll up subagent (child) sessions into session token stats
## Environment

- CCHV: v1.27.0 (headless server mode, `cchv-server --serve`)
- Provider: OpenCode 1.18.29, SQLite storage (`~/.local/share/opencode/opencode.db`)
- OS: Ubuntu (WSL2)

## Summary

The Token Statistics page (`messages.tokenStats.existing`) reports a session's cost/tokens from **that session's own messages only**. When a session delegates work to subagents (Task tool), those runs are stored as separate child sessions, so their billed usage is never added to the parent. There is no "this session including its subagents" total anywhere in the UI, which means the true cost of one originating session is understated and has to be summed by hand.

## Current behavior

- `get_session_token_stats` → `load_stats_messages(provider, &session_path)` loads only the requested session (`src-tauri/src/commands/stats.rs`), so descendant sessions are never aggregated.
- OpenCode: a Task run is a child row in `session` whose `parent_id` points at its parent, and it is loaded independently by `src-tauri/src/providers/opencode.rs`. Its messages carry `is_sidechain: None`, so the `billing_total` / `conversation_only` sidechain switch described in `docs/specs/token-stats-consistency-plan.md` does not cover OpenCode child sessions at all.
- The frontend only prices the single session's `model_distribution` (`calculateGlobalCostSummary` in `src/components/TokenStatsViewer.tsx`), so the "Current Session" card shows the main thread only.
- Child sessions currently appear as ordinary rows in the project session list (`load_sessions_from_db` does not filter on `parent_id`), so the project-level total happens to include them — but there is no per-parent roll-up and the per-session card excludes them.

Concrete example (real rows from an OpenCode DB; `own` = `session.cost`, `descendants` = recursive sum over `parent_id`):

| Session | own cost | descendant cost | true total |
|---|---|---|---|
| 深圳北京往返机票优惠方案调研 | $4.91 | $12.54 | $17.45 |
| 小模型SFT配置校验实验调研 | $4.68 | $14.91 | $19.58 |
| Obsidian Vault 技术文章写作规范 | $2.21 | $7.33 | $9.54 |

The data is all stored — the global `opencode stats` total (~$671 across 247 sessions) already includes subagent sessions — only the per-session roll-up is missing.

## Expected behavior

Add an explicit roll-up that optionally includes subagent / child sessions, consistently across session → project → global.

Suggested approach:

1. In `get_session_token_stats`, when the provider supports child sessions, compute both:
   - `session` = conversation only (main thread)
   - `session_with_subagents` = main thread + all descendant sessions, recursively (a subagent may itself spawn subagents).
2. Reuse the existing parent→child linkage added for #560 (`providers::opencode::load_child_sessions`).
3. Surface it on the Token Statistics page, e.g. `Session (main) / Subagents / Session (with subagents)`, matching the existing `Billing Total = Conversation Only + Sidechain` breakdown.
4. Apply the same inclusion rule to the project and global aggregates so totals reconcile (the invariant already stated in `docs/specs/token-stats-consistency-plan.md`: global == sum(project), project == sum(session)).
5. Default should follow the existing policy — billed usage is included by default (`billing_total`) — with an explicit opt-out.

## Why this matters

Delegated runs can dominate cost. In my data subagents are ~10% globally, but for heavily delegating sessions they are frequently more than half of the true total (see table above). Users budgeting per session currently cannot see that cost without manually summing child sessions.

## Notes

- Distinct from #560 (viewing subagent conversations) and #414 / #321 (listing skill / subagent usage); this is specifically about **cost/token aggregation**.
- Related: #150 (sidechain accounting) and `docs/specs/token-stats-consistency-plan.md`.
- Happy to help test or provide the recursive SQL used for the numbers above.

GoldenStain:
## What & why

Closes #577.

OpenCode persists a subagent (`Task`) run as its own row in the `session` table, linked by `session.parent_id`. A session's token stats only read that session's own messages, so the billed usage of everything it delegated was missing from the Token Statistics page — the "Current Session" card understated the true cost, often by more than half for heavily delegating sessions.

This adds an optional rolled-up `subagent_stats` (all descendant sessions, recursively) to `SessionTokenStats`, and shows it on the Token Statistics page.

**Backend**
- `providers/opencode.rs`: new `parse_session_path` and a recursive `load_descendant_sessions_from_db` that walks children, grandchildren, … (a subagent can itself spawn subagents). `load_descendant_sessions` is the default-root wrapper.
- `commands/stats.rs`: `get_session_token_stats` attaches `subagent_stats` for `OpenCode` in billing mode. The roll-up reuses the existing `build_session_token_stats_from_messages`, so token extraction and dedup rules match a normal session.
- Conversation-only mode stays main-thread only, and providers that keep subagents inline (Claude's `isSidechain`) are skipped, so their billed usage is not counted twice.

**Frontend**
- New `SubagentRollupCard` rendered in the Current Session section when `subagent_stats` has descendants: subagent tokens/cost/session count, plus a combined main + subagent total.
- `SessionTokenStats` TS type extended; i18n keys added to all five locales.

Example (real OpenCode data):

| Session | own cost | descendants | true total |
|---|---|---|---|
| 深圳北京往返机票优惠方案调研 | $4.91 | $12.54 | $17.45 |
| Obsidian Vault 技术文章写作规范 | $2.21 | $7.33 | $9.54 |

## Type

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / perf
- [ ] Docs

## Checklist

  - Rust: `sqlite_load_descendant_sessions_walks_recursively`, `parse_session_path_extracts_project_and_session`, `test_subagent_stats_from_messages_aggregates_descendants`, `test_build_subagent_token_stats_is_opencode_only`
- [x] **i18n**: new strings are `t()`-wrapped and present in all 5 languages — `pnpm run i18n:validate` passes

N/A — no new command. The optional field rides on the existing `get_session_token_stats`, so both the Tauri and the Axum WebUI paths serialize it unchanged.

## Verification

- `cargo fmt --all -- --check` ✅
- `cargo clippy --all-targets -- -D warnings` ✅
- `cargo clippy --all-targets --all-features -- -D warnings` ✅
- `cargo test -- --test-threads=1` ✅ (1009 passed)
- `cargo test --features webui-server -- --test-threads=1` ✅ (1055 passed)
- `pnpm exec tsc --build .` ✅ · `pnpm lint` ✅ · `pnpm exec vitest run` ✅ (1199 passed; one unrelated pre-existing 5s timeout in `RecentEditsOptionsMenu.test.tsx` that passes when given more time)
- `pnpm run i18n:validate` ✅

## Summary by CodeRabbit

- **New Features**
  - Analytics now displays usage from delegated subagent sessions, including token totals, estimated costs, session counts, and combined session totals.
  - Subagent usage is rolled up across descendant sessions for supported billing analytics.
  - Added translations for subagent analytics in English, Japanese, Korean, Simplified Chinese, and Traditional Chinese.

- **Bug Fixes**
  - Improved accuracy of billing statistics by including eligible delegated-session usage while keeping conversation-only statistics focused on the main session.

- **Tests**
  - Added coverage for descendant-session aggregation, counts, provider handling, and session path validation.


jhlee0409 (changes requested):
Thanks for this — the writeup and the #577 example numbers made it easy to review, and scoping the roll-up to OpenCode only (avoiding a double-count with Claude's inline `isSidechain` messages) is exactly right. I checked that the descendant messages can't collide with the parent's dedup keys and that this doesn't touch the Claude stats cache path.

Two things before merge:

1. **`session_count` can drift from the shown totals under a date filter.** `build_subagent_token_stats` sets `session_count` to the full, unfiltered descendant count, but `message_count`/tokens are filtered by `s_limit`/`e_limit` inside `build_session_token_stats_from_messages`. Since the Token Statistics date picker feeds straight into `get_session_token_stats`, a filtered view can show e.g. "5 subagent sessions" with tokens from only 1 of them. Could you derive `session_count` from the distinct `session_id`s that actually pass the date filter?
2. **Cycle guard on the recursive CTE.** `load_descendant_sessions_from_db` uses `UNION ALL`, which won't terminate if `parent_id` ever contains a cycle (corrupted DB, partial write). Switching to plain `UNION` bounds it via row dedup — cheap insurance since we don't control the OpenCode DB's integrity.

Scope question: #577 also asks for the same roll-up at project/global level so totals reconcile end-to-end (global == sum(project) == sum(session)); this PR covers the single-session view only. Do you plan to follow up with the project/global part, or should this PR be treated as the full fix for #577? Just want it to be a deliberate choice.

CI: `Rust Tests`/`Frontend Tests` were at `action_required` (fork gate) — I've approved them. Locally on this branch: 1009 Rust tests, clippy, fmt, tsc, and i18n:validate all passed.

PR #579, opened 2026-09-10
fix(vite): set worker.format to es so the search worker bundles on macOS

louisje:
## Problem

`pnpm exec vite build` fails on macOS x86_64 (verified on a clean `main` checkout, Node 22 and 24):

```
[vite:worker-import-meta-url] Invalid value "iife" for option "worker.format"
- UMD and IIFE output formats are not supported for code-splitting builds.
file: src/utils/searchIndex.ts
```

The search worker (`src/utils/searchWorker.ts`, introduced in #352) is instantiated with `{ type: "module" }`, but Vite's default `worker.format` is `iife`. FlexSearch's dependency graph contains dynamic imports, which rollup on macOS treats as code-splitting — and IIFE output cannot represent that. CI (ubuntu) does not hit this because rollup's platform-specific behavior differs, so `frontend-tests.yml` (tsc + vitest only) never catches it.

## Fix

Set `worker: { format: "es" }` in `vite.config.ts`. ES output matches the worker's `type: "module"` instantiation and works on every platform.

## Verification

- `pnpm exec vite build` succeeds on macOS x86_64 (was: hard error)
- `pnpm tsc --build .` and `pnpm -s vitest run` unchanged (1200 passed)


jhlee0409:
Thanks for tracking this down — good catch that CI (ubuntu, tsc/vitest only) can't see this class of bug at all.

I rebuilt both `develop` and this branch with `pnpm exec vite build`: the change flips the search worker's output from an IIFE wrapper to a plain ES module, matching the `{ type: "module" }` instantiation in `searchIndex.ts`. It's the only Web Worker in the codebase, and `just frontend-build` (used by both `tauri-build` and the WebUI embed) shares this same `vite.config.ts`, so nothing else is affected. I couldn't reproduce the failure itself on an arm64 Mac, which is consistent with it being machine-specific.

I pushed one small follow-up commit (13e9c88): inside the async `defineConfig` callback the object literal isn't contextually typed, so `format: "es"` widened to `string` and `tsc --build` failed with TS2769. `format: "es" as const` fixes it — tsc and `vite build` both pass. The remaining `Frontend Tests` failure (`RecentEditsOptionsMenu › reports a filter change` timing out) is pre-existing on develop and unrelated to this change, so I've merged.

One small nit for later: the comment attributes the failure to "rollup on macOS treat[ing] it as code-splitting", but rollup's chunking isn't platform-specific — it's more likely a dependency-resolution difference in the worker's graph.

PR #416, opened 2026-06-21
feat(providers): add Continue.dev, PearAI, Kilo Code, Goose, Crush, and llm

jhlee0409:
## Summary

Expands supported AI-coding assistants from **14 to 19 provider IDs** (+ Kilo Code folded into the Cline-family reader) — **6 new tools**, from the platform-coverage survey.

| Provider | Storage | Approach |
|---|---|---|
| **Continue.dev** | `~/.continue/sessions/*.json` | new JSON module, grouped by `workspaceDirectory`; honors `CONTINUE_GLOBAL_DIR` |
| **PearAI** | `~/.pearai/sessions/*.json` | shares a parameterized Continue-family core (`Family`) with `continue_dev` |
| **Kilo Code** | globalStorage `kilocode.kilo-code` | one-line add to the Cline `EXTENSIONS` table (per-task files byte-identical to Cline/Roo) |
| **Goose** | `sessions.db` (SQLite) | grouped by `working_dir`; maps Goose `MessageContent` → Claude-style blocks |
| **Crush** | per-project `./.crush/crush.db` (SQLite) | aider-style discovery scan; maps Crush `parts` → Claude-style blocks |
| **llm** | `io.datasette.llm/logs.db` (SQLite) | single synthetic project; responses → user+assistant turns w/ token usage |

## Verification

- All on-disk schemas (and Goose/Crush message-content shapes) were confirmed from each tool's **open-source code** (migrations / message types).
- SQLite readers are structured as `*_conn(&Connection)` and **unit-tested against in-memory databases**; JSON readers use synthetic fixtures.
- Gates: `tsc` ✅ · `vitest` 864 ✅ · `eslint` ✅ · `i18n:validate` ✅ · `cargo test` 719 ✅ · `clippy -D warnings` ✅ · `fmt` ✅

## Notes / limitations

- ⚠️ **Not yet runtime-verified against real installs** — a real-file sanity check per tool is recommended before release (timestamp units / content edge cases).
- **Crush** DBs are per-project (no global root) → Crush is not tracked by the file watcher or the WebUI session-path allowlist. Desktop IPC works; the WebUI headless server cannot read scattered per-project DBs.
- `rusqlite` was already a dependency; no new crates added.

## Deferred (follow-ups)

- **Amazon Q CLI** — needs its own reader (table `conversations`, not the `conversations_v2` the `kiro` provider queries).
- **Qwen Code** — `logs.json` stores user messages only; assistant content lives in `checkpoint-*.json` → needs design.
[...]

PR #580, opened 2026-09-10
feat(providers): support new Kilo Code OpenCode-core store

louisje:
## Motivation

Kilo Code left the Cline/Roo fork lineage in May 2026 and now runs on the OpenCode core. Verified on a live install:

- Sessions live in a drizzle-managed SQLite database (`~/.local/share/kilo/kilo.db`) whose schema and `message`/`part` JSON shapes are identical to OpenCode's.
- Kilo's own legacy migration moves the old VS Code `globalState` task index into the new store (`"kilo.legacyMigrationStatus": "completed"`; the old `globalStorage/kilocode.kilo-code` task dir is emptied the same day `kilo.db`'s earliest session begins, 2026-05-27). Reading the new store therefore covers the full history.
- The current reader (added in #416, fixed in #422/#425) reads the old store only, so migrated installs show zero Kilo sessions.

## Design

- **Dedicated `kilo` provider** reading `~/.local/share/kilo/kilo.db`, reusing the OpenCode reader with a parameterized db file. `opencode://` paths are rebranded to `kilo://` so the two stores can never cross-reference ids. `KILO_HOME` / `XDG_DATA_HOME` overrides mirror `opencode.rs`.
- **The pre-migration Cline-family reader stays untouched** in `cline.rs` as a fallback for installations that never ran Kilo's migration (migrated installs keep no task index there, so the two readers do not double-report).
- Wired into: provider detection, scan/load/search routing, file watcher, subagent child sessions (`kilo://` scheme), stats, recent edits, and the WebUI path guard.
- Frontend: `ProviderId`, provider list/translations/badges, project tree tab counts, i18n keys (5 locales), README updates (5 locales).

## Verification

- `cargo test -- --test-threads=1`: **1028 passed / 0 failed** (rebased on develop)
- `cargo clippy --all-targets -- -D warnings` and `--all-features`: clean; `cargo fmt --check`: clean
- `pnpm tsc --build .`: clean; `pnpm -s vitest run`: **1200 passed**; `pnpm run i18n:validate`: pass
- Live-store smoke test against a real `kilo.db`: 25 projects, 338 messages in the newest session, provider stamped `kilo`

Note: depends on #579 for `vite build` on macOS (unrelated worker-format fix, kept separate per review preference).

## Summary by CodeRabbit

- **New Features**
  - Added Kilo Code as a supported provider.
  - Automatically discovers Kilo projects and loads sessions, messages, child sessions, and statistics.
  - Added Kilo Code to global search and project navigation.
  - Supports Kilo session data stored in its standard local data directory.
- **Localization**
  - Added Kilo Code provider labels in English, Japanese, Korean, Simplified Chinese, and Traditional Chinese.
- **Bug Fixes**
  - Improved handling and logging of provider search failures.


jhlee0409 (changes requested):
Thanks for the deep dive on Kilo's migration to the OpenCode core, and for reusing the `opencode.rs` reader via the parameterized `_store`/`_at` functions instead of duplicating it — that's the right pattern. Coexistence with the pre-migration Cline-family Kilo reader from #422 also looks correctly designed (no double-counting).

Two things need fixing before merge (both still present at `93d0433`):

1. **Live file-watch refresh doesn't fire for Kilo.** `collect_watch_paths` in `lib.rs` correctly adds the Kilo store root, but `extract_provider_paths` / `extract_opencode_db_event` in `commands/watcher.rs` only recognize filenames starting with `"opencode."`. `kilo.db` / `kilo.db-wal` changes fall through to `None`, so no `FileWatchEvent` is emitted — new Kilo sessions/messages only show up after a manual rescan. Please add a `kilo.`-aware branch (mirroring `extract_opencode_db_event`, emitting `kilo://…`) plus a test.
2. **Global stats summary omits Kilo.** `StatsProvider::Kilo` is wired into per-project/session stats, but `get_global_stats_summary` in `commands/stats.rs` never includes it in its per-provider collection — the same gap CodeRabbit flagged. Please add Kilo there (mirroring the OpenCode block) with a regression test using a `KILO_HOME` fixture.

Smaller items, not blocking:
- The new Kilo watch-path block in `lib.rs` uses `.is_file()` (follows symlinks) before registering the store root; the `kimi_code` block just above shows the hardened `symlink_metadata` pattern.
- A couple of error strings in the shared `opencode.rs` reader (`OpenCodeProjectRef::parse`, `require_absolute_path(..., "OpenCode base path")`) still say "OpenCode" for Kilo failures.
- `kilo::detect()` checks for a `storage/` dir that Kilo doesn't have (carried over from `opencode::detect()`).
- `SOURCE_COST_ONLY_PROVIDER_IDS` in the analytics calculations lists `opencode` but not `kilo`, and parallel-task grouping in `messageCategories.ts` is hardcoded to `opencode` — Kilo won't get either.

CI: `Rust Tests` and `Frontend Tests` were at `action_required` (fork gate); I've approved them on the current head. #579 (the vite worker fix this depends on for local macOS builds) is being merged now.

louisje:
Thanks for the thorough review — all points addressed in `c7fedcf`.

**Blocking**

1. **Live file-watch for Kilo** — `extract_opencode_db_event` in `commands/watcher.rs` now recognises `kilo.db` / `kilo.db-wal` and emits a `kilo://*` wildcard (same shape as `opencode://*`, so `watcherSlice.resolveWildcard` already handles it). Tests: `test_extract_kilo_db_event_emits_kilo_wildcard`, plus coverage for the existing `opencode.*` branch and an unknown `.db` falling through to `None`.
2. **Global stats summary** — `get_global_stats_summary` now collects `StatsProvider::Kilo` right after the OpenCode block. Regression test `test_global_summary_includes_kilo_store` builds a `kilo.db` fixture under `KILO_HOME` (via a new `kilo::test_support` module shared with the provider's own tests) and asserts projects/sessions/token distribution/`provider_distribution` all include Kilo.

**Non-blocking**

- `lib.rs`: the Kilo watch root is registered only when both the root dir and `kilo.db` are real (non-symlink) via `symlink_metadata`, mirroring the `kimi_code` block.
- Error strings: rather than threading a display name through every shared `opencode.rs` function, `kilo.rs` restates the shared reader's errors (`OpenCode` → `Kilo Code`, `opencode://` → `kilo://`) in a single `rebrand_error` applied to every delegated call. Test `errors_are_reported_in_kilo_terms` covers project path, session path and base-path failures. Happy to switch to parameterising the reader if you'd prefer that.
- `kilo::detect()` no longer checks for a `storage/` dir; availability is `kilo.db` alone.
- Frontend: `kilo` added to `SOURCE_COST_ONLY_PROVIDER_IDS` and the parallel-task collector (same `Task` tool as OpenCode), each with a test.

Verified locally: `cargo test -- --test-threads=1` (watcher/stats/kilo/opencode modules, 120 passed), `cargo clippy -D warnings`, `cargo fmt --check`, `pnpm tsc --build .`, eslint, and vitest on the touched suites.

